PR #7, opened 2020-03-10
Bootstrap

xp-1000:
This PR aims to bootstrap this repository translating 4 modules:
- filter-tags common module
- nginx, kong and system

The goal is to:
- keep every features we had on counterpart datadog repository: https://github.com/claranet/terraform-datadog-monitors
- smartly implement new features depending of signalfx (provider) capabilities
- rethink some things now we begin directly in terraform 0.12 version for example

Here we need to approve the implementation for the 3 detectors modules (nginx, kong and system) focusing on detectors code, default behavior and customization possibilities.

For now, you can ignore README which will be generated automatically bu CI.
But we need to merge this first to be able to create the CI and continue to translate other modules from the same way of those in this PR (if it seems good for you).

I will also create a documentation of convention how to create new modules (as for datadog in the past) when this PR will be merged (and so the way it works validated).


common/filter-tags/README.md
```diff
@@ -0,0 +1,47 @@
+# FILTER TAGS Datadog Generator
+
+## How to use this module
+
+This module usage should be transparent because it should be used inside each detectors modules directly.
+Here is a simple example but it is advisable to see how are created other existing detectors modules:
+
+```
+module "filter-tags" {
+  source = "../../common/filter-tags"
```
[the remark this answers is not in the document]
xp-1000: no it depends where your module is located in repo so I cannot provide a generic answer to that.

do you prefer I use a repo link instead of a local address ?

common/filter-tags/outputs.tf
```diff
@@ -0,0 +1,85 @@
+output "filter_custom" {
+  description = "The full filtering pattern to use in detectors"
+  value = "${var.filter_use_defaults == true ? var.filter_defaults :
```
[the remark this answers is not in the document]
xp-1000: as it contains interpolation synthax we need the `"${..}"`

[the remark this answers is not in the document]
pdecat: String interpolation is not needed with HCL2 in this context. See [](url)

common/filter-tags/outputs.tf
```diff
@@ -0,0 +1,85 @@
...
+                    " ",
+                    ""
+                  )
+                )
+              )
+            )
+          )
+      )}"
+    }"
+  }"
```
pdecat:
```suggestion
  value = (
    var.filter_use_defaults == true ?
    var.filter_defaults
    :
    var.filter_custom_includes == "" ?
    join(
      " and ",
      formatlist(
        "(not filter('%s'))",
        split(
          ";",
          replace(
            replace(
              var.filter_custom_excludes,
              ":",
              "', '"
            ),
            " ",
            ""
          )
        )
      )
    )
    :
    var.filter_custom_excludes == "" ?
    join(
      " and ",
      formatlist(
        "(not filter('%s'))",
        split(
          ";",
          replace(
            replace(
              var.filter_custom_includes,
              ":",
              "', '"
            ),
            " ",
            ""
          )
        )
      )
    )
    :
    join(
      " and ",
      list(
        join(
          " and ",
          formatlist(
            "filter('%s')",
            split(
              ";",
              replace(
                replace(
                  var.filter_custom_includes,
                  ":",
                  "', '"
                ),
                " ",
                ""
              )
            )
          )
        ),
        join(
          " and ",
          formatlist(
            "(not filter('%s'))",
            split(
              ";",
              replace(
                replace(
                  var.filter_custom_excludes,
                  ":",
                  "', '"
                ),
                " ",
                ""
              )
            )
          )
        )
      )
    )
  )
```

middleware/kong/detectors-kong.tf
```diff
@@ -0,0 +1,49 @@
+resource "signalfx_detector" "heartbeat" {
+  name = "${upper(join("", formatlist("[%s]", var.prefixes_slug)))}[${upper(var.environment)}] Kong heartbeat"
```
pdecat: I'd remove the `upper()` function calls to leave more control on the resulting detector name to the user of the modules. I find that especially desired since these are publicly released.

middleware/kong/README.md
```diff
@@ -0,0 +1,57 @@
...
+| detectors\_disabled | Disable all detectors in this module | `bool` | `false` | no |
+| environment | Infrastructure environment | `string` | n/a | yes |
+| filter\_custom\_excludes | Tags to exclude when using custom filtering (i.e "tag1:val1;tag2:val2") | `string` | `""` | no |
+| filter\_custom\_includes | Tags to filter signals on when custom filtering is used (i.e "tag1:val1;tag2:val2") | `string` | `""` | no |
+| filter\_use\_defaults | Use default filtering which follows tagging convention | `bool` | `true` | no |
+| heartbeat\_disabled | Disable all alerting rules for heartbeat detector | `bool` | n/a | yes |
+| heartbeat\_notifications | Notification recipients semicolon for every alerting rules of heartbeat detector | `string` | `""` | no |
+| heartbeat\_timeframe | Timeframe for system not reporting detector (i.e. "10m") | `string` | `"20m"` | no |
+| notifications | Notification recipients semicolon separated (i.e. "Email,my@mail.com;PagerDuty,credentialId") | `string` | n/a | yes |
+| prefixes\_slug | Prefixes list to prepend between brackets on every monitors names before environment | `list` | `[]` | no |
```
pdecat: Using _slug_ here feels wrong, that [word](https://en.wikipedia.org/wiki/Slug_(publishing)) is overused in this context IMO.

middleware/kong/README.md
```diff
@@ -0,0 +1,57 @@
...
+| treatment\_limit\_aggregation\_function | Aggregation function and group by for treatment limit detector (i.e. ".mean(by=['host']).") | `string` | `""` | no |
+| treatment\_limit\_disabled | Disable all alerting rules for treatment limit detector | `bool` | n/a | yes |
+| treatment\_limit\_disabled\_critical | Disable critical alerting rule for treatment limit detector | `bool` | n/a | yes |
+| treatment\_limit\_disabled\_warning | Disable warning alerting rule for treatment limit detector | `bool` | n/a | yes |
+| treatment\_limit\_notifications | Notification recipients semicolon for every alerting rules of treatment limit detector | `string` | `""` | no |
```
pdecat: Why not pass notification recipients as strings lists instead of performing string concatenation / splitting everywhere?

common/filter-tags/outputs.tf
xp-1000: oh my bad, I did a wrong test and missing to replace the negative part .. thanks!

middleware/kong/detectors-kong.tf
xp-1000: it was to preserve / enforce a consistency but ok it makes sens

middleware/kong/README.md
xp-1000: it is only because I did not know `coalescelist` function and did not have the time to search instantly. after I forgot so thanks for that

common/filter-tags/outputs.tf
```diff
@@ -0,0 +1,89 @@
...
+                " ",
+                ""
+              )
+            )
+          )
+        )
+      )
+    )
+  )
+}
```
pdecat: As discussed, simpler variant (with passed as lists):
```suggestion
locals {
  filter_custom_includes = join(
    " and ",
    formatlist(
      "(filter('%s'))",
      [for tag in var.filter_custom_includes : replace(replace(tag, ":", "', '"), " ", "")]
    )
  )

  filter_custom_excludes = join(
    " and ",
    formatlist(
      "(not filter('%s'))",
      [for tag in var.filter_custom_excludes : replace(replace(tag, ":", "', '"), " ", "")]
    )
  )
}

output "filter_custom" {
  description = "The full filtering pattern to use in detectors"
  value = coalesce(
    join(
      " and ",
      compact(
        [
          local.filter_custom_includes,
          local.filter_custom_excludes,
        ]
      )
    ),
    var.filter_defaults
  )
}
```

xp-1000: It is awesome ! I tested it works fine, thank you

PR #18, opened 2020-04-28
Cloud gcp

mel-splunk:
Review complete.   Replaced forecast detectors with resource running out and anomaly detector with sudden change as discussed

### DIFF
--- a/cloud/gcp/gce/instance/modules.tf
+++ b/cloud/gcp/gce/instance/modules.tf
@@ -1,0 +1,7 @@
+module "filter-tags" {
+  source = "github.com/claranet/terraform-signalfx-detectors.git//common/filter-tags"
+
+  filter_defaults        = "filter('gcp_metadata_sfx_env', '${var.environment}') and filter('gcp_metadata_sfx_monitored', 'true')"
+  filter_custom_includes = var.filter_custom_includes
+  filter_custom_excludes = var.filter_custom_excludes
+}
--- a/cloud/gcp/pubsub/subscription/detectors-subscription.tf
+++ b/cloud/gcp/pubsub/subscription/detectors-subscription.tf
@@ -1,0 +1,75 @@
+resource "signalfx_detector" "heartbeat" {
+  name = "${join("", formatlist("[%s]", var.prefixes))}[${var.environment}] GCP Pub/Sub Subscription heartbeat"
+
+  program_text = <<-EOF
+    from signalfx.detectors.not_reporting import not_reporting
+    signal = data('subscription/pull_request_count', filter=${module.filter-tags.filter_custom}).publish('signal')
+    not_reporting.detector(stream=signal, resource_identifier=None, duration='${var.heartbeat_timeframe}').publish('CRIT')
+EOF
+
+  rule {
+    description           = "has not reported in ${var.heartbeat_timeframe}"
+    severity              = "Critical"
+    detect_label          = "CRIT"
+    disabled              = coalesce(var.heartbeat_disabled, var.detectors_disabled)
+    notifications         = coalescelist(var.heartbeat_notifications, var.notifications)
+    parameterized_subject = "[{{ruleSeverity}}]{{{detectorName}}} {{{readableRule}}} on {{{dimensions}}}"
+  }
+}
+
+resource "signalfx_detector" "oldest_unacked_message" {
+  name = "${join("", formatlist("[%s]", var.prefixes))}[${var.environment}] GCP Pub/Sub Subscription oldest unacknowledged message"
+
+  program_text = <<-EOF
+    signal = data('subscription/oldest_unacked_message_age', filter=filter('monitored_resource', 'pubsub_subscription') and ${module.filter-tags.filter_custom})${var.oldest_unacked_message_aggregation_function}${var.oldest_unacked_message_transformation_function}.publish('signal')
+    detect(when(signal >= ${var.oldest_unacked_message_threshold_critical})).publish('CRIT')
+    detect(when(signal >= ${var.oldest_unacked_message_threshold_warning}) and when(signal < ${var.oldest_unacked_message_threshold_critical})).publish('WARN')
+EOF
+
+  rule {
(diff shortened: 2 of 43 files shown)


cloud/gcp/gce/instance/modules.tf
```diff
@@ -0,0 +1,7 @@
+module "filter-tags" {
+  source = "github.com/claranet/terraform-signalfx-detectors.git//common/filter-tags"
+
+  filter_defaults        = "filter('gcp_tag_env', '${var.environment}') and filter('gcp_tag_sfx_monitored', 'true')"
```
pdecat, line 4: As instance _labels_ (_tags_ are a network feature in GCP domain) are not collected by SignalFx's GCP integration right now, we can rely on instance _metadata_ to filter metrics:
```suggestion
  filter_defaults        = "filter('gcp_metadata_sfx_env', '${var.environment}') and filter('gcp_metadata_sfx_monitored', 'true')"
```

# Notes
* As there's a high chance of having conflicts on the `env` metadata name, I suggest prefixing it with `sfx_` like it's done for `sfx_monitored`.

* To be available, such _metadata_ has to be both:
  - added at the [instance (or project) level](https://cloud.google.com/compute/docs/storing-retrieving-metadata#custom), e.g.:
    - `gcloud compute instances add-metadata myinstance --zone=europe-west1-c --metadata sfx_env=true`
    - `gcloud compute instances add-metadata myinstance --zone=europe-west1-c --metadata sfx_monitored=true`
  - whitelisted at the SignalFx's GCP integration level (requires [SignalFx terraform provider v4.22.0+](https://github.com/terraform-providers/terraform-provider-signalfx/blob/v4.22.0/CHANGELOG.md#4220-may-29-2020), see https://docs.signalfx.com/en/latest/integrations/google-cloud-platform.html#compute-engine-instance).

WDYT?

cc @xp-1000

cloud/gcp/pubsub/subscription/variables.tf
```diff
@@ -0,0 +1,279 @@
...
+}
+
+variable "push_latency_transformation_window" {
+  description = "Transformation window for push_latency detector (i.e. 5m, 20m, 1h, 1d)"
+  type        = string
+  default     = "10m"
+}
+
+variable "push_latency_threshold_critical" {
+  description = "Critical threshold for push_latency detector"
```
rsicart, line 184: Metric unit for `subscription/push_request_latencies` on [google api](https://cloud.google.com/monitoring/api/metrics_gcp#gcp-pubsub) is "microseconds".
Shouldn't be the threshold 5000000 instead of 5000?

Same for warning threshold.

cloud/gcp/pubsub/subscription/detectors-subscription.tf
```diff
@@ -0,0 +1,96 @@
...
+	rule {
+		description           = "at >= ${var.push_latency_anomaly_threshold_critical}"
+		severity              = "Critical"
+		detect_label          = "CRIT"
+		disabled              = coalesce(var.push_latency_anomaly_disabled_critical, var.push_latency_anomaly_disabled, var.detectors_disabled)
+		notifications         = coalescelist(var.push_latency_anomaly_notifications_critical, var.push_latency_anomaly_notifications, var.notifications)
+		parameterized_subject = "[{{ruleSeverity}}]{{{detectorName}}} {{{readableRule}}} ({{inputs.signal.value}}) on {{{dimensions}}}"
+	}
+
+}
```
rsicart, line 96: What do you think about adding more detectors, as mentioned in [google docs](https://cloud.google.com/pubsub/docs/monitoring)?

Some detectors for:

* `subscription/num_undelivered_messages`
* `subscription/pull_ack_message_operation_count`, filtered by `response_code != "success"`
* `subscription/push_request_count`, filtered by `response_code != "success"`
* `subscription/push_request_count`
* `subscription/num_outstanding_messages`
* `subscription/num_undelivered_messages`

I don't know if it's too much...

cloud/gcp/pubsub/topic/detectors-topics.tf
```diff
@@ -0,0 +1,91 @@
...
+	rule {
+		description           = "is too high >= ${var.unavailable_sending_operations_ratio_threshold_warning}"
+		severity              = "Warning"
+		detect_label          = "WARN"
+		disabled              = coalesce(var.unavailable_sending_operations_ratio_disabled_warning, var.unavailable_sending_operations_ratio_disabled, var.detectors_disabled)
+		notifications         = coalescelist(var.unavailable_sending_operations_ratio_notifications_warning, var.unavailable_sending_operations_ratio_notifications, var.notifications)
+		parameterized_subject = "[{{ruleSeverity}}]{{{detectorName}}} {{{readableRule}}} ({{inputs.signal.value}}) on {{{dimensions}}}"
+	}
+
+}
```
rsicart, line 91: As mentioned also in [google docs](https://cloud.google.com/pubsub/docs/monitoring#keeping_publishers_healthy), to keep publishers healthy, a detector for:
* `topic/send_request_count`, grouped by `response_code`

cloud/gcp/pubsub/subscription/detectors-subscription.tf
xp-1000: yes you are right. that said I think we should merge the PR with the current scope which works and add them in another PR. for now the priority is to translate "smartly" everything from https://github.com/claranet/terraform-signalfx-detectors.

I added an issue this in our bugtracker to not forget.

cloud/gcp/pubsub/topic/detectors-topics.tf
xp-1000: added to issue !

cloud/gcp/pubsub/subscription/variables.tf
xp-1000: good catch ! the error was also in the original implementation : https://github.com/claranet/terraform-datadog-monitors/blob/master/cloud/gcp/pubsub/subscription/inputs.tf#L120

xp-1000 (approved):
Seems good for a first release now

PR #24, opened 2020-05-22
Add supervisor detector

cvauvarin:
Add `supervisor` detector now that the monitor is available in agent v5.2.

The default behavior is to :
* Alert CRITICAL when the process is not running and not stopped
* Alert WARNING when the process has been stopped manually

### DIFF
--- a/middleware/supervisor/README.md
+++ b/middleware/supervisor/README.md
@@ -53,4 +53,5 @@ Creates SignalFx detectors with the following checks:
 
 ## Related documentation
 
+[Official documentation for supervisor](https://docs.signalfx.com/en/latest/integrations/agent/monitors/supervisor.html)
 [Supervisor state definition](http://supervisord.org/subprocess.html#process-states)
--- a/middleware/supervisor/README.md
+++ b/middleware/supervisor/README.md
@@ -1,0 +1,56 @@
+# MIDDLEWARE SUPERVISOR SignalFx detectors
+
+## How to use this module
+
+```hcl
+module "signalfx-detectors-middleware-supervisor" {
+  source      = "github.com/claranet/terraform-signalfx-detectors.git//middleware/supervisor?ref={revision}"
+
+  environment = var.environment
+  notifications = var.notifications
+}
+
+```
+
+## Purpose
+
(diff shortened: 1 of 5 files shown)


middleware/supervisor/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Outputs
+
+| Name | Description |
+|------|-------------|
+| heartbeat\_id | id for detector heartbeat |
+| process\_state\_id | id for detector process state |
+
+## Related documentation
+
```
xp-1000, line 55: may be could we add https://docs.signalfx.com/en/latest/integrations/agent/monitors/supervisor.html ?

middleware/supervisor/detectors-supervisor.tf
```diff
@@ -0,0 +1,47 @@
...
+    parameterized_subject = "[{{ruleSeverity}}]{{{detectorName}}} {{{readableRule}}} on {{{dimensions}}}"
+  }
+}
+
+resource "signalfx_detector" "process_state" {
+  name = "${join("", formatlist("[%s]", var.prefixes))}[${var.environment}] Supervisor process"
+
+  program_text = <<-EOF
+		signal = data('supervisor.state', filter=${module.filter-tags.filter_custom})${var. process_state_aggregation_function}.${var. process_state_transformation_function}(over='${var.process_state_transformation_window}').publish('signal')
+		detect(when(signal > ${var.process_state_threshold_critical})).publish('CRIT')
```
xp-1000, line 25: this way could lead to undesired alerts no ?

for example STOPPING or STARTING states will trigger alerts which seem not legitimate.

of course we could hope process_state_transformation_window is big enough to avoid that but it seems to be a better approach to use custom detector and alerts only if one of undesired state : https://docs.signalfx.com/en/latest/detect-alert/alert-condition-reference/custom-threshold.html

I suggest to map:

- STARTING and STARTED as OK
- UNKNOWN and STOPPING as warning  
- every other states as critical

cvauvarin: As discussed I think we can merge it that way and we will improve it if needed later.

PR #25, opened 2020-06-10
Add RabbitMQ detectors

cvauvarin:
Add multiple detectors for RabbitMQ : 
* RabbitMQ Node vm_memory usage
* RabbitMQ Node sockets usage
* RabbitMQ Node file descriptors usage
* RabbitMQ Node process usage
* RabbitMQ Queue messages ready
* RabbitMQ Queue messages unacknowledged (disabled by default)
* RabbitMQ Queue messages ack rate (disabled by default)
* RabbitMQ Queue consumer utilisation (disable by default)

The behavior is different than usual for the last three as it is possible to create multiple detectors with different filters (if you need to have very specific threshold for some queue or vhost). Ready to discuss any thought on this implementation.


middleware/rabbitmq/detectors-rabbitmq.tf
```diff
@@ -0,0 +1,256 @@
...
+
+resource "signalfx_detector" "file_descriptors" {
+  name = "${join("", formatlist("[%s]", var.prefixes))}[${var.environment}] RabbitMQ Node file descriptors usage"
+
+  program_text = <<-EOF
+        A = data('gauge.node.fd_used', filter=filter('plugin', 'rabbitmq') and ${module.filter-tags.filter_custom})${var.file_descriptors_aggregation_function}
+        B = data('gauge.node.fd_total', filter=filter('plugin', 'rabbitmq') and ${module.filter-tags.filter_custom})${var.file_descriptors_aggregation_function}
+        signal = (A/B).scale(100).${var.file_descriptors_transformation_function}(over='${var.file_descriptors_transformation_window}').publish('signal')
+        detect(when(signal > ${var.file_descriptors_threshold_critical})).publish('CRIT')
+        detect(when(signal > ${var.file_descriptors_threshold_warning})).publish('WARN')
```
xp-1000, line 28: you should use a "negative condition" to clean other thresholds.
```suggestion
        detect(when(signal > ${var.file_descriptors_threshold_warning}) and when(signal <= ${var.file_descriptors_threshold_critical})).publish('WARN')
        ```
```

middleware/rabbitmq/detectors-rabbitmq.tf
```diff
@@ -0,0 +1,256 @@
...
+
+resource "signalfx_detector" "processes" {
+  name = "${join("", formatlist("[%s]", var.prefixes))}[${var.environment}] RabbitMQ Node process usage"
+
+  program_text = <<-EOF
+        A = data('gauge.node.proc_used', filter=filter('plugin', 'rabbitmq') and ${module.filter-tags.filter_custom})${var.processes_aggregation_function}
+        B = data('gauge.node.proc_total', filter=filter('plugin', 'rabbitmq') and ${module.filter-tags.filter_custom})${var.processes_aggregation_function}
+        signal = (A/B).scale(100).${var.processes_transformation_function}(over='${var.processes_transformation_window}').publish('signal')
+        detect(when(signal > ${var.processes_threshold_critical})).publish('CRIT')
+        detect(when(signal > ${var.processes_threshold_warning})).publish('WARN')
```
xp-1000, line 58:
```suggestion
        detect(when(signal > ${var.processes_threshold_warning}) and when(signal <= ${var.processes_threshold_critical})).publish('WARN')
```

middleware/rabbitmq/detectors-rabbitmq.tf
```diff
@@ -0,0 +1,256 @@
...
+
+resource "signalfx_detector" "sockets" {
+  name = "${join("", formatlist("[%s]", var.prefixes))}[${var.environment}] RabbitMQ Node sockets usage"
+
+  program_text = <<-EOF
+        A = data('gauge.node.sockets_used', filter=filter('plugin', 'rabbitmq') and ${module.filter-tags.filter_custom})${var.sockets_aggregation_function}
+        B = data('gauge.node.sockets_total', filter=filter('plugin', 'rabbitmq') and ${module.filter-tags.filter_custom})${var.sockets_aggregation_function}
+        signal = (A/B).scale(100).${var.sockets_transformation_function}(over='${var.sockets_transformation_window}').publish('signal')
+        detect(when(signal > ${var.sockets_threshold_critical})).publish('CRIT')
+        detect(when(signal > ${var.sockets_threshold_warning})).publish('WARN')
```
xp-1000, line 88:
```suggestion
        detect(when(signal > ${var.sockets_threshold_warning}) and when(signal < ${var.sockets_threshold_critical})).publish('WARN')
```

middleware/rabbitmq/detectors-rabbitmq.tf
```diff
@@ -0,0 +1,256 @@
...
+
+resource "signalfx_detector" "vm_memory" {
+  name = "${join("", formatlist("[%s]", var.prefixes))}[${var.environment}] RabbitMQ Node vm_memory usage"
+
+  program_text = <<-EOF
+        A = data('gauge.node.mem_used', filter=filter('plugin', 'rabbitmq') and ${module.filter-tags.filter_custom})${var.vm_memory_aggregation_function}
+        B = data('gauge.node.mem_limit', filter=filter('plugin', 'rabbitmq') and ${module.filter-tags.filter_custom})${var.vm_memory_aggregation_function}
+        signal = (A/B).scale(100).${var.vm_memory_transformation_function}(over='${var.vm_memory_transformation_window}').publish('signal')
+        detect(when(signal > ${var.vm_memory_threshold_critical})).publish('CRIT')
+        detect(when(signal > ${var.vm_memory_threshold_warning})).publish('WARN')
```
xp-1000, line 118:
```suggestion
        detect(when(signal > ${var.vm_memory_threshold_warning}) and when(signal < ${var.vm_memory_threshold_critical})).publish('WARN')
```

middleware/rabbitmq/detectors-rabbitmq.tf
```diff
@@ -0,0 +1,256 @@
...
+  EOF
+
+  rule {
+    description           = "is approaching the watermark"
```
xp-1000, line 122: watermark -> limit?

middleware/rabbitmq/detectors-rabbitmq.tf
```diff
@@ -0,0 +1,256 @@
...
+    severity              = "Warning"
+    detect_label          = "WARN"
+    disabled              = coalesce(var.vm_memory_disabled_warning, var.vm_memory_disabled, var.detectors_disabled)
+    notifications         = coalescelist(var.vm_memory_notifications_warning, var.vm_memory_notifications, var.notifications)
+    parameterized_subject = "[{{ruleSeverity}}]{{{detectorName}}} {{{readableRule}}} ({{inputs.signal.value}}) on {{{dimensions}}}"
+  }
+}
+
+resource "signalfx_detector" "messages_ready" {
+  for_each = var.messages_ready_thresholds
```
xp-1000, line 141: why not using an aggregation mechanism, at least by default, to monitor every queues with the same threshold instead of having to know and configure every queues manually.

Also, I understand if this is usual to configure each queue manually with custom threshold for each one this could be painful but the "normal" way to do is :
- having a full generic detector which will works out the box
- make it possible to multiple instance the terraform module (with different filters and detectors disabled)

so here you could: 
- have a generic detector to monitor every queues by default.
- import this module once with default config and disable this detector.
- import this module for each queue you want to monitor enabling only this detector with its custom threshold.

does it make sens?

middleware/rabbitmq/detectors-rabbitmq.tf
cvauvarin: Yes, this is the name in RabbitMQ configuration (`vm_memory_high_watermark`) : https://www.rabbitmq.com/memory.html

middleware/rabbitmq/detectors-rabbitmq.tf
cvauvarin: > why not using an aggregation mechanism, at least by default, to monitor every queues with the same threshold instead of having to know and configure every queues manually.

By default it will trigger an alert for each queue, but you are right I will add an aggregation mechanism as this can lead to duplicate alert in a cluster configuration.

> does it make sens?

I thought about it and found really unpractical to call the entire module each time you want to add a custom configuration for a dedicated queue or vhost. As each detector is still created but only the alerting rules are disabled this can lead to many many unused detectors in the UI.

There are various use case where you want different thresholds and behavior and it is really unlikely that you want the same for all your queues, for example : 
* Queues with lots of messages from time to time (at night for example) where you only want to check whether your consumer are working properly : in this case you don't really care about checking if there is too many messages in the queue
* Queues where you need to consume your message really fast and you don't want too many messages in ready state : you might consider checking the ack rate and the ready message threshold

The `for_each` method might not be the right choice, maybe we can also consider to have two modules (this is what I was considering at first) : 
* `rabbitmq/node`: will contains the detectors related to the middleware itself and the various system check we need (memory, fd, sockets, process), we can add some useful cluster detectors in the future also,
* `rabbitmq/queue`: will contains the detectors related to the broker usage itself and the customer usage of it. Detectors create by this one will in most cases require some finetuning as it depends on the use case

What to you think ?

tchernomax: For my part I found the for_each approach pretty convenient.

I don't understand your sentence 
> why not using an aggregation mechanism, at least by default, to monitor every queues with the same threshold instead of having to know and configure every queues manually.

currently, with messages_ack_rate_thresholds = 
```
    "default" = {
      filter             = "filter('name', '*')"
      threshold_critical = "1/60"
      threshold_warning  = "2/60"
    }
```
every queues are monitored with the same threshold (maybe I misunderstood the signalflow).

middleware/rabbitmq/detectors-rabbitmq.tf
xp-1000: I don't say it is not correct but it seems more understandable in my opinion and this is just as relevant as watermark while the metric is named `gauge.node.mem_limit`

middleware/rabbitmq/detectors-rabbitmq.tf
xp-1000: > As each detector is still created but only the alerting rules are disabled this can lead to many many unused detectors in the UI.

It is a very good point. I confess I had in mind datadog implementation which use count on the terraform resource but here you are right.

splitting module seems to be the best approach. with terraform 0.13 we can for each on module so it will be possible to implement to loop in the "client" terraform side importing multiple times the module with different thresholds config.

from this way we can preserve the generic implementation and a straightforward usage in future.

are you ok?

cvauvarin: Seems to be a good compromise, just changed with the module splited in two.

PR #27, opened 2020-06-15
Add Haproxy detectors

cvauvarin:
Add Haproxy detectors :

* Haproxy heartbeat
* Haproxy server status
* Haproxy backend status
* Haproxy session limit (on frontend, backend and server)
* Haproxy 4xx response rate (on frontend and backend with http mode)
* Haproxy 5xx response rate (on frontend and backend with http mode)

https://github.com/signalfx/signalfx-agent/pull/1357 is needed to merge this as `haproxy_session_limit` was not reported by the agent.

### DIFF
--- a/middleware/haproxy/detectors-haproxy.tf
+++ b/middleware/haproxy/detectors-haproxy.tf
@@ -1,0 +1,144 @@
+resource "signalfx_detector" "heartbeat" {
+  name = "${join("", formatlist("[%s]", var.prefixes))}[${var.environment}] Haproxy heartbeat"
+
+  program_text = <<-EOF
+		from signalfx.detectors.not_reporting import not_reporting
+		signal = data('haproxy_session_current', filter=filter('aws_state', 'running') and filter('gcp_status', '*RUNNING}') and filter('azure_power_state', 'PowerState/running') and ${module.filter-tags.filter_custom}).publish('signal')
+		not_reporting.detector(stream=signal, resource_identifier=['host'], duration='${var.heartbeat_timeframe}').publish('CRIT')
+  EOF
+
+  rule {
+    description           = "has not reported in ${var.heartbeat_timeframe}"
+    severity              = "Critical"
+    detect_label          = "CRIT"
+    disabled              = coalesce(var.heartbeat_disabled, var.detectors_disabled)
+    notifications         = coalescelist(var.heartbeat_notifications, var.notifications)
+    parameterized_subject = "[{{ruleSeverity}}]{{{detectorName}}} {{{readableRule}}} on {{{dimensions}}}"
+  }
+}
+
+resource "signalfx_detector" "server_status" {
+  name = "${join("", formatlist("[%s]", var.prefixes))}[${var.environment}] Haproxy server status"
+
+  program_text = <<-EOF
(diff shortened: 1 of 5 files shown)


middleware/haproxy/detectors-haproxy.tf
```diff
@@ -0,0 +1,144 @@
+resource "signalfx_detector" "heartbeat" {
+  name = "${join("", formatlist("[%s]", var.prefixes))}[${var.environment}] Haproxy heartbeat"
+
+  program_text = <<-EOF
+		from signalfx.detectors.not_reporting import not_reporting
+		signal = data('haproxy_session_current', filter=filter('aws_state', 'running') and filter('gcp_status', '*RUNNING}') and filter('azure_power_state', 'PowerState/running') and ${module.filter-tags.filter_custom}).publish('signal')
```
xp-1000, line 6: this metric seems always reported but may be it is more "natural" to use the `haproxy_status` ?

middleware/haproxy/detectors-haproxy.tf
```diff
@@ -0,0 +1,144 @@
...
+    notifications         = coalescelist(var.heartbeat_notifications, var.notifications)
+    parameterized_subject = "[{{ruleSeverity}}]{{{detectorName}}} {{{readableRule}}} on {{{dimensions}}}"
+  }
+}
+
+resource "signalfx_detector" "server_status" {
+  name = "${join("", formatlist("[%s]", var.prefixes))}[${var.environment}] Haproxy server status"
+
+  program_text = <<-EOF
+		signal = data('haproxy_status', filter=filter('type', '2') and ${module.filter-tags.filter_custom})${var.server_status_aggregation_function}.${var.server_status_transformation_function}(over='${var.server_status_transformation_window}').publish('signal')
```
xp-1000, line 24: I think `1` should be hardcoded because changing its value does not make sens (and will be a risk to "break" the detector)

middleware/haproxy/detectors-haproxy.tf
```diff
@@ -0,0 +1,144 @@
...
+    notifications         = coalescelist(var.server_status_notifications_critical, var.server_status_notifications, var.notifications)
+    parameterized_subject = "[{{ruleSeverity}}]{{{detectorName}}} {{{readableRule}}} ({{inputs.signal.value}}) on {{{dimensions}}}"
+  }
+}
+
+resource "signalfx_detector" "backend_status" {
+  name = "${join("", formatlist("[%s]", var.prefixes))}[${var.environment}] Haproxy backend status"
+
+  program_text = <<-EOF
+		signal = data('haproxy_status', filter=filter('type', '1') and ${module.filter-tags.filter_custom})${var.backend_status_aggregation_function}.${var.backend_status_transformation_function}(over='${var.backend_status_transformation_window}').publish('signal')
```
xp-1000, line 42: I think `1` should be hardcoded because changing its value does not make sens (and will be a risk to "break" the detector)

middleware/haproxy/detectors-haproxy.tf
cvauvarin: Yes that what I planned to choose at first but it is not a default metrics so I thought it was better to take this one which is a default one.

PR #29, opened 2020-06-17
Processes

aurelien-claranet:
Check processes, 2 possible workaround:
- with the warning threshold, to check if the number of a single process is matching.
- without the warning threshold, to check if the processes are present

### DIFF
--- a/system/processes/README.md
+++ b/system/processes/README.md
@@ -1,4 +1,4 @@
-# SYSTEM GENERIC SignalFx detectors
+# SYSTEM PROCESSES SignalFx detectors
 
 ## How to use this module
 
@@ -25,14 +25,14 @@ Creates SignalFx detectors with the following checks:
 | processes\_aggregation\_function | Aggregation function and group by for processes detector (i.e. ".mean(by=['host']).") | `string` | `""` | no |
 | processes\_disabled | Disable all alerting rules for processes detector | `bool` | n/a | yes |
 | processes\_disabled\_critical | Disable critical alerting rule for processes detector | `bool` | n/a | yes |
-| processes\_disabled\_warning | Disable warning alerting rule for processes detector | `bool` | n/a | yes |
+| processes\_disabled\_warning | Disable warning alerting rule for processes detector | `true` | n/a | yes |
 | processes\_notifications | Notification recipients list for every alerting rules of processes detector | `list` | `[]` | no |
 | processes\_notifications\_critical | Notification recipients list for critical alerting rule of processes detector | `list` | `[]` | no |
 | processes\_notifications\_warning | Notification recipients list for warning alerting rule of processes detector | `list` | `[]` | no |
 | processes\_threshold\_critical | Critical threshold for processes detector | `number` | `1` | no |
-| processes\_threshold\_warning | Warning threshold for processes detector | `number` | `2` | no |
+| processes\_threshold\_warning | Warning threshold for processes detector | `number` | `1` | no |
 | processes\_transformation\_function | Transformation function for processes detector (mean, min, max) | `string` | `"min"` | no |
-| processes\_transformation\_window | Transformation window for processes detector (i.e. 5m, 20m, 1h, 1d) | `string` | `"1h"` | no |
+| processes\_transformation\_window | Transformation window for processes detector (i.e. 5m, 20m, 1h, 1d) | `string` | `"15min"` | no |
 | detectors\_disabled | Disable all detectors in this module | `bool` | `false` | no |
 | environment | Infrastructure environment | `string` | n/a | yes |
 | filter\_custom\_excludes | List of tags to exclude when custom filtering is used | `list` | `[]` | no |
@@ -52,3 +52,8 @@ Creates SignalFx detectors with the following checks:
 
 ## Notes
 
+By default, this detectors will check if a process is present or not. 
+
+The "warning" state can be enable if you need to check if a process need a minimal amount of process running.
+Ex: I need 5 "foo" processes for my api to run correctly, so I will put in warning 5 and in critical 1. 
+In that case, "warning" will raise if I’m under 5 processes and "critical" if no process is found.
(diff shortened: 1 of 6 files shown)


system/processes/detectors-processes.tf
```diff
@@ -0,0 +1,30 @@
+resource "signalfx_detector" "processes" {
+  name = "${join("", formatlist("[%s]", var.prefixes))}[${var.environment}] Processes aliveness"
+
+  program_text = <<-EOF
+        signal = data('ps_count.processes', filter=${module.filter-tags.filter_custom})${var.processes_aggregation_function}.${var.processes_transformation_function}(over='${var.processes_transformation_window}').publish('signal')
+        detect(when(signal < ${var.processes_threshold_critical})).publish('CRIT')
+        detect(when((signal < ${var.processes_threshold_warning}) and (signal > ${var.processes_threshold_critical}))).publish('WARN')
```
xp-1000, line 7:
```suggestion
        detect(when((signal < ${var.processes_threshold_warning}) and when(signal >= ${var.processes_threshold_critical})).publish('WARN')
```

system/processes/detectors-processes.tf
```diff
@@ -0,0 +1,30 @@
...
+  EOF
+
+  rule {
+    description           = "is too low < ${var.processes_threshold_critical}"
```
xp-1000, line 11:
```suggestion
    description           = "count is too low < ${var.processes_threshold_critical}"
```

"processes count is" or "processes are", as you prefer

system/processes/detectors-processes.tf
```diff
@@ -0,0 +1,30 @@
...
+    severity              = "Critical"
+    detect_label          = "CRIT"
+    disabled              = coalesce(var.processes_disabled_critical, var.processes_disabled, var.detectors_disabled)
+    notifications         = coalescelist(var.processes_notifications_critical, var.processes_notifications, var.notifications)
+    parameterized_subject = "[{{ruleSeverity}}]{{{detectorName}}} {{{readableRule}}} ({{inputs.signal.value}}) on {{{dimensions}}}"
+  }
+
+  rule {
+    description           = "is too low < ${var.processes_threshold_warning}"
```
xp-1000, line 20:
```suggestion
    description           = "count is too low < ${var.processes_threshold_warning}"
```

system/processes/README.md
```diff
@@ -0,0 +1,54 @@
+# SYSTEM GENERIC SignalFx detectors
```
xp-1000, line 1:
```suggestion
# SYSTEM PROCESSES SignalFx detectors
```

system/processes/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+| Name | Description |
+|------|-------------|
+| processes\_id | id for detector processes |
+
+## Related documentation
+
+[Official documentation for processes](https://docs.signalfx.com/en/latest/integrations/agent/monitors/collectd-processes.html)
+
+## Notes
```
xp-1000, line 53: could be deleted if not useful but do not hesitate to provide a configuration sample ;)

system/processes/variables.tf
```diff
@@ -0,0 +1,106 @@
...
+variable "processes_transformation_function" {
+  description = "Transformation function for processes detector (mean, min, max)"
+  type        = string
+  default     = "min"
+}
+
+variable "processes_transformation_window" {
+  description = "Transformation window for processes detector (i.e. 5m, 20m, 1h, 1d)"
+  type        = string
+  default     = "1h"
```
xp-1000, line 92: seems very long no?

system/processes/variables.tf
```diff
@@ -0,0 +1,106 @@
...
+variable "processes_threshold_critical" {
+  description = "Critical threshold for processes detector"
+  type        = number
+  default     = 1
+}
+
+variable "processes_threshold_warning" {
+  description = "Warning threshold for processes detector"
+  type        = number
+  default     = 1
```
xp-1000, line 104: I am not sure how you think the behavior of this detector but it seems not natural to have same threshold for warning and critical.
I think we should choose a default behavior (like critical if no process at all) and may be make it possible to change behavior to another (like warning if the number of processes do not match).
But the default one should be generic and "plug and play". 
It is possible to add into "Notes" of readme if there is a "custom way" to use.

system/processes/detectors-processes.tf
aurelien-claranet: Change done

system/processes/detectors-processes.tf
aurelien-claranet: Change done

system/processes/detectors-processes.tf
aurelien-claranet: Change done

system/processes/README.md
aurelien-claranet: I always missed it, 
Change done

system/processes/variables.tf
aurelien-claranet: hmmm, so, at first I was not thinking about adding a warning, it will not be used very often, only if the customer need to check one of his own process and need to have at least x processes running.
By default, we will use this detector, with warning in disable state, to monitor processes like sshd, cron, rsyslogd.
So I think I have to put this:

```
53 variable "processes_disabled_warning" {
 54   description = "Disable warning alerting rule for processes detector"
 55   type        = bool
 56   default     = null
 57 }
```
like:

`56   default     = true`

We are speaking about our experience in ex-oxalide, if you see others implementations of this detector, feel free to discuss it with me.

That’s why we used "collectd processes" and not "telegraf processstat" that, for me, is too verbose and with too many options not needed.

system/processes/README.md
aurelien-claranet: I fill it up

system/processes/variables.tf
aurelien-claranet: changed to 15min

PR #30, opened 2020-06-22
Nagios script status check

cvauvarin:
A simple detector to alert with the same behavior as Nagios.

### DIFF
--- a/system/nagios-status-check/README.md
+++ b/system/nagios-status-check/README.md
@@ -1,0 +1,77 @@
+# SYSTEM Nagios status check SignalFx detectors
+
+## How to use this module
+
+```hcl
+module "signalfx-detectors-nagios-status-check" {
+  source      = "github.com/claranet/terraform-signalfx-detectors.git//system/nagios-status-check?ref={revision}"
+
+  environment = var.environment
+  notifications = var.notifications
+}
+
+```
+
+## Purpose
+
+Creates SignalFx detectors with the following checks:
+
+- Nagios check status
+
+This module has been designed to alert with the same behavior as [Nagios](https://nagios-plugins.org/doc/guidelines.html#AEN78), basically a gauge which is equal to 1 will trigger à `WARNING` and equal to 2 a `CRITICAL`. 
+
+On the agent side, you need to send your metrics with the [telegraf exec plugin][https://docs.signalfx.com/en/latest/integrations/agent/monitors/telegraf-exec.html] which includes a [parser for nagios](https://github.com/influxdata/telegraf/blob/master/plugins/parsers/nagios/parser.go). The metric is named `nagios_state.state`, you need to add an `extraDimensions` to your monitor in order to be able to differantiate multiple script states.
+
+Here is an example : 
+
+```yaml
+- type: telegraf/exec
+  intervalSeconds: 180
+  command: /usr/local/bin/scripts/check-ipmitool.pl
+  extraDimensions:
+    script: check_ipmitool
+  telegrafParser:
+    dataFormat: nagios
+- type: telegraf/exec
+  intervalSeconds: 900  
+  command: /usr/local/bin/scripts/check-megaraidsas
+  extraDimensions:
+    script: check_megaraidsas
+  telegrafParser:
+    dataFormat: nagios
+```
+
+## Inputs
+
+| Name | Description | Type | Default | Required |
+|------|-------------|:----:|:-----:|:-----:|
+| detectors\_disabled | Disable all detectors in this module | bool | `"false"` | no |
+| environment | Infrastructure environment | string | n/a | yes |
+| filter\_custom\_excludes | List of tags to exclude when custom filtering is used | list | `[]` | no |
+| filter\_custom\_includes | List of tags to include when custom filtering is used | list | `[]` | no |
+| notifications | Notification recipients list for every detectors | list | n/a | yes |
+| prefixes | Prefixes list to prepend between brackets on every monitors names before environment | list | `[]` | no |
+| status\_check\_aggregation\_function | Aggregation function and group by for status\_check detector \(i.e. ".mean\(by=\['host'\]\)."\) | string | `""` | no |
+| status\_check\_disabled | Disable all alerting rules for status\_check detector | bool | `"null"` | no |
+| status\_check\_disabled\_critical | Disable critical alerting rule for status\_check detector | bool | `"null"` | no |
+| status\_check\_disabled\_unknown | Disable unknown alerting rule for status\_check detector | bool | `"null"` | no |
+| status\_check\_disabled\_warning | Disable warning alerting rule for status\_check detector | bool | `"null"` | no |
+| status\_check\_lasting\_duration\_seconds | Minimum duration that conditions must be true before raising alert \(in seconds\) | string | `"900"` | no |
+| status\_check\_notifications | Notification recipients list for every alerting rules of status\_check detector | list | `[]` | no |
+| status\_check\_notifications\_critical | Notification recipients list for critical alerting rule of status\_check detector | list | `[]` | no |
+| status\_check\_notifications\_unknown | Notification recipients list for unknown alerting rule of status\_check detector | list | `[]` | no |
+| status\_check\_notifications\_warning | Notification recipients list for warning alerting rule of status\_check detector | list | `[]` | no |
+| status\_check\_transformation\_function | Transformation function for status\_check detector \(i.e. ".mean\(over='5m'\)"\) | string | `""` | no |
+
+## Outputs
+
+| Name | Description |
+|------|-------------|
+| status\_check\_id | id for detector status\_check |
+
+## Related documentation
+
+- [Official documentation for the telegraf exec plugin](https://docs.signalfx.com/en/latest/integrations/agent/monitors/telegraf-exec.html)
+- [Nagios documentation](https://nagios-plugins.org/doc/guidelines.html#AEN78)
+- [Telegraf exec plugin](https://github.com/influxdata/telegraf/tree/master/plugins/inputs/exec)
+- [Telegraf nagios parser](https://github.com/influxdata/telegraf/tree/master/plugins/parsers/nagios)
--- a/system/nagios-status-check/detectors-nagios-status-check.tf
+++ b/system/nagios-status-check/detectors-nagios-status-check.tf
@@ -1,0 +1,37 @@
+resource "signalfx_detector" "status_check" {
+  name = "${join("", formatlist("[%s]", var.prefixes))}[${var.environment}] Nagios check status"
+
+  program_text = <<-EOF
+        signal = data('nagios_state.state', filter=${module.filter-tags.filter_custom})${var.status_check_aggregation_function}${var.status_check_transformation_function}.publish('signal')
+        detect(when(signal == 1, lasting='${var.status_check_lasting_duration_seconds}s')).publish('WARN')
+        detect(when(signal == 2, lasting='${var.status_check_lasting_duration_seconds}s')).publish('CRIT')
+        detect(when(signal == 3, lasting='${var.status_check_lasting_duration_seconds}s')).publish('UNKN')
+  EOF
+
+  rule {
+    description           = "is UNKNOWN, please check the script output on the host"
+    severity              = "Critical"
+    detect_label          = "UNKN"
+    disabled              = coalesce(var.status_check_disabled_unknown, var.status_check_disabled, var.detectors_disabled)
+    notifications         = coalescelist(var.status_check_notifications_unknown, var.status_check_notifications, var.notifications)
+    parameterized_subject = "[{{ruleSeverity}}]{{{detectorName}}} {{{readableRule}}} ({{inputs.signal.value}}) on {{{dimensions}}}"
+  }
+
+  rule {
+    description           = "is CRITICAL, please check the script output on the host"
+    severity              = "Critical"
+    detect_label          = "CRIT"
+    disabled              = coalesce(var.status_check_disabled_critical, var.status_check_disabled, var.detectors_disabled)
+    notifications         = coalescelist(var.status_check_notifications_critical, var.status_check_notifications, var.notifications)
+    parameterized_subject = "[{{ruleSeverity}}]{{{detectorName}}} {{{readableRule}}} ({{inputs.signal.value}}) on {{{dimensions}}}"
+  }
+
+  rule {
+    description           = "is WARNING, please check the script output on the host"
+    severity              = "Warning"
+    detect_label          = "WARN"
+    disabled              = coalesce(var.status_check_disabled_warning, var.status_check_disabled, var.detectors_disabled)
+    notifications         = coalescelist(var.status_check_notifications_warning, var.status_check_notifications, var.notifications)
+    parameterized_subject = "[{{ruleSeverity}}]{{{detectorName}}} {{{readableRule}}} ({{inputs.signal.value}}) on {{{dimensions}}}"
+  }
+}
--- a/system/nagios-status-check/variables.tf
+++ b/system/nagios-status-check/variables.tf
@@ -1,0 +1,105 @@
+# Global
+
+variable "environment" {
+  description = "Infrastructure environment"
+  type        = string
+}
+
+# SignalFx Module specific
+
+variable "notifications" {
+  description = "Notification recipients list for every detectors"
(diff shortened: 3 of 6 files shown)


system/nagios-status-check/detectors-nagios-status-check.tf
```diff
@@ -0,0 +1,27 @@
+resource "signalfx_detector" "status_check" {
+  name = "${join("", formatlist("[%s]", var.prefixes))}[${var.environment}] Custom script status"
+
+  program_text = <<-EOF
+        signal = data('gauge.status', filter=filter('plugin', 'nagios') and ${module.filter-tags.filter_custom})${var.status_check_aggregation_function}.${var.status_check_transformation_function}(over='${var.status_check_transformation_window}').publish('signal')
+        detect(when(signal >= 1) and when(signal < 2)).publish('WARN')
+        detect(when(signal >= 2) and when(signal < 3)).publish('CRIT')
```
tchernomax, line 7:
```suggestion
        detect(when(signal >= threshold(1)) and when(signal < 2)).publish('WARN')
        detect(when(signal >= threshold(2)) and when(signal < 3)).publish('CRIT')
```

tchernomax: And for UNKNOWN ?

Maybe you should write:
```suggestion
        detect(when(signal >= threshold(1)) and when(signal < 2)).publish('WARN')
        detect(when(signal >= threshold(2))).publish('CRIT')
```
Or create un UNKNOWN dedicated rule.

cvauvarin: I don't know if it's worse having an UNKNOWN state as we do not have it in any other detectors. I was also thinking to not put an upper threshold to the CRITICAL state as if the script is really not working we can consider that we must be alerted.

system/nagios-status-check/README.md
```diff
@@ -0,0 +1,69 @@
...
+	out=$(exec $@)
+	if [[ $out =~ "OK" ]]; then
+		value=${STATUS["OK"]}
+	elif [[ $out =~ "WARNING" ]]; then
+		value=${STATUS["WARNING"]}
+	elif [[ $out =~ "CRITICAL" ]]; then
+		value=${STATUS["CRITICAL"]}
+	else 
+		value=${STATUS["UNKNOWN"]}
+	fi
```
tchernomax, line 59: https://gitlab.oxalide.net/bu-saas/signalfx/ansible-signalfx-conf/-/merge_requests/1/diffs#diff-content-d2d8d64aadbd93a3b9906908ecd864cf7a6c90c9

Du coup on peut peut-être modifier le script ici aussi.

system/nagios-status-check/README.md
```diff
@@ -0,0 +1,69 @@
...
+- Custom script status
+
+This module has been designed to alert with the same behavior as [Nagios](https://nagios-plugins.org/doc/guidelines.html#AEN78), basically a gauge which is equal to 1 will trigger à `WARNING` and equal to 2 a `CRITICAL`. 
+
+On the agent side, you need to send your metrics with a metric named `gauge.status` in SignalFx, the `plugin_instance` dimension will allow you to find which script is failing. The easiest way to do that is to use the [exec collectd plugin](https://docs.signalfx.com/en/latest/integrations/agent/monitors/collectd-custom.html).
+
+Here is an example : 
+
+```yaml
+- type: collectd/custom
```
xp-1000, line 28: may be using https://docs.signalfx.com/en/latest/integrations/agent/monitors/telegraf-exec.html could ease the configuration (at least avoid the xml template ..)

system/nagios-status-check/README.md
```diff
@@ -0,0 +1,69 @@
...
+  template: |
+    LoadPlugin exec
+    <Plugin exec>
+       Exec "signalfx-agent" "/usr/local/bin/check-generic-status-nagios-plugin.sh" "/usr/local/bin/scripts/custom-nagios-script.pl"
+    </Plugin>
+```
+
+The following script (`check-generic-status-nagios-plugin.sh` in the example above) takes a script path as an argument, execute it, check the output and send the status in the collectd format.
+
+```bash
```
xp-1000, line 39: why not put this script as file in this directory and forward to it from this readme ?

system/nagios-status-check/README.md
```diff
@@ -0,0 +1,69 @@
...
+#!/bin/bash
+
```
xp-1000, line 41: `set -euo pipefail` should pass like a charm while your script is well written :)

system/nagios-status-check/README.md
```diff
@@ -0,0 +1,69 @@
...
+INTERVAL="${COLLECTD_INTERVAL:-10}"
+HOSTNAME="${COLLECTD_HOSTNAME:-`hostname -f`}"
+
+declare -A STATUS
+STATUS=( ["OK"]=0 ["WARNING"]=1 ["CRITICAL"]=2 ["UNKNOWN"]=3 )
+
+while sleep "$INTERVAL"
+do
+	out=$(exec $@)
```
xp-1000, line 50: why not rely on exit code instead of output ? I know lot of scripts are not very respectful of the convention and could send `CRIT` or even `Critical` instead of `CRITICAL` or even in some cases do not put this at the start of its output ..

if exit code is not possible I would make this more tolerant matching the first "short" version of the pattern without case so basically `crit`

what do you think?

system/nagios-status-check/detectors-nagios-status-check.tf
```diff
@@ -0,0 +1,27 @@
+resource "signalfx_detector" "status_check" {
```
xp-1000, line 1: could be done in another future PR but could be useful to also add an heartbeat detector.

In this case, I am pretty sure we will have to cover both cases (erratic or regular reporting) so we could eventually add a boolean flag parameter to the wrapper script and add a dimension "regular" or wathever which will allow to filter only on those checks from the hearbeat detector.

From this way:
- we could be alerted of not reporting checks which should always do
- but allow the manage the exception for erratic checks just by adding the right parameter to the wrapper script in the monitor conf

system/nagios-status-check/detectors-nagios-status-check.tf
```diff
@@ -0,0 +1,27 @@
...
+  name = "${join("", formatlist("[%s]", var.prefixes))}[${var.environment}] Custom script status"
```
xp-1000, line 2: I think you should add the "Nagios" context in the name of this detector. even if it is fully correct this module and their detectors are dedicated to the nagios logic, maybe : 

```suggestion
  name = "${join("", formatlist("[%s]", var.prefixes))}[${var.environment}] Nagios check status"
```

cvauvarin: Done

system/nagios-status-check/README.md
cvauvarin: This seems to be a much better solution, I changed the PR.

system/nagios-status-check/README.md
cvauvarin: Not needed any more with `telegraf/exec`.

system/nagios-status-check/README.md
cvauvarin: Not needed any more with `telegraf/exec`.

system/nagios-status-check/README.md
cvauvarin: Not needed any more with `telegraf/exec` which has exactly this behavior.

cvauvarin:
@xp-1000 and @tchernomax I changed the PR in order to use the `telegraf/exec` plugin instead. @tchernomax the only drawbacks is that the `telegraf` parser only checks for the exit code and not the output of the script but I think we can rewrite the few script we have that do not use the Nagios convention.

system/nagios-status-check/README.md
```diff
@@ -0,0 +1,70 @@
...
+- type: telegraf/exec
+  intervalSeconds: 30
+  command: /usr/local/bin/scripts/check-ipmitool.pl
+  extraDimensions:
+    script: check_ipmitool
+  telegrafParser:
+    dataFormat: nagios
+```
+
+Ideally, use an `extraDimensions` which allows to easily identify the executed script.
```
xp-1000, line 37: I think you can remove this line while you already say it just above : 

> you need to add an `extraDimensions` to your monitor in order to be able to differantiate multiple script states.

and I prefer "you need" than "ideally" while this is a requirement in my mind.

may be can you provide 1 config example with 2 examples instead of only one to show the diff between dimensions.

system/nagios-status-check/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Outputs
+
+| Name | Description |
+|------|-------------|
+| status\_check\_id | id for detector status\_check |
+
+## Related documentation
+
+[Official documentation for the telegraf exec plugin](https://docs.signalfx.com/en/latest/integrations/agent/monitors/telegraf-exec.html)
```
xp-1000, line 67: you should probable add newline or create a list, every docs will be on the same line which is difficult to read

system/nagios-status-check/detectors-nagios-status-check.tf
```diff
@@ -0,0 +1,27 @@
...
+  name = "${join("", formatlist("[%s]", var.prefixes))}[${var.environment}] Nagios check status"
+
+  program_text = <<-EOF
+        signal = data('nagios_state.state', filter=${module.filter-tags.filter_custom})${var.status_check_aggregation_function}.${var.status_check_transformation_function}(over='${var.status_check_transformation_window}').publish('signal')
+        detect(when(signal >= threshold(1)) and when(signal < 2)).publish('WARN')
+        detect(when(signal >= threshold(2))).publish('CRIT')
+  EOF
+
+  rule {
+    description           = "is not OK, please check the script output on the host"
```
xp-1000, line 11: this is true but what about say explicitely it is a critical to respect the nagios terminology (even if you binded it to the same severity from signalfx)

system/nagios-status-check/detectors-nagios-status-check.tf
```diff
@@ -0,0 +1,27 @@
+resource "signalfx_detector" "status_check" {
+  name = "${join("", formatlist("[%s]", var.prefixes))}[${var.environment}] Nagios check status"
+
+  program_text = <<-EOF
+        signal = data('nagios_state.state', filter=${module.filter-tags.filter_custom})${var.status_check_aggregation_function}.${var.status_check_transformation_function}(over='${var.status_check_transformation_window}').publish('signal')
+        detect(when(signal >= threshold(1)) and when(signal < 2)).publish('WARN')
+        detect(when(signal >= threshold(2))).publish('CRIT')
```
xp-1000, line 7: is there no "Unknown" status ?

system/nagios-status-check/detectors-nagios-status-check.tf
```diff
@@ -0,0 +1,27 @@
+resource "signalfx_detector" "status_check" {
+  name = "${join("", formatlist("[%s]", var.prefixes))}[${var.environment}] Nagios check status"
+
+  program_text = <<-EOF
+        signal = data('nagios_state.state', filter=${module.filter-tags.filter_custom})${var.status_check_aggregation_function}.${var.status_check_transformation_function}(over='${var.status_check_transformation_window}').publish('signal')
```
xp-1000, line 5: I udpated other modules code to merge transformation function into only one variable like for the aggregation

```suggestion
        signal = data('nagios_state.state', filter=${module.filter-tags.filter_custom})${var.status_check_aggregation_function}${var.status_check_transformation_function}.publish('signal')
```

system/nagios-status-check/variables.tf
```diff
@@ -0,0 +1,93 @@
...
+variable "status_check_aggregation_function" {
+  description = "Aggregation function and group by for status_check detector (i.e. \".mean(by=['host']).\")"
+  type        = string
+  default     = ""
+}
+
+variable "status_check_transformation_function" {
+  description = "Transformation function for status_check detector (mean, min, max)"
+  type        = string
+  default     = "min"
```
xp-1000, line 86:
```suggestion
  default     = ".min(over='15m')"
```

xp-1000 (changes requested):
also please could you rebase to enjoy last CI

system/nagios-status-check/detectors-nagios-status-check.tf
cvauvarin: Yes way better, done.

system/nagios-status-check/detectors-nagios-status-check.tf
cvauvarin: Yes there is, I was wondering if it worth it but I just added it with the severity "Critical" as there is no unknown in SignalFx.

system/nagios-status-check/detectors-nagios-status-check.tf
```diff
@@ -0,0 +1,37 @@
+resource "signalfx_detector" "status_check" {
+  name = "${join("", formatlist("[%s]", var.prefixes))}[${var.environment}] Nagios check status"
+
+  program_text = <<-EOF
+        signal = data('nagios_state.state', filter=${module.filter-tags.filter_custom})${var.status_check_aggregation_function}${var.status_check_transformation_function}.publish('signal')
+        detect(when(signal >= threshold(1)) and when(signal < 2)).publish('WARN')
```
xp-1000, line 6:
```suggestion
        detect(when(signal == threshold(1))).publish('WARN')
```

should be enough and easier and will allow lasting function

system/nagios-status-check/detectors-nagios-status-check.tf
```diff
@@ -0,0 +1,37 @@
...
+        detect(when(signal >= threshold(2)) and when(signal < 3)).publish('CRIT')
+        detect(when(signal >= threshold(3))).publish('UNKN')
+  EOF
+
+  rule {
+    description           = "is UNKNOWN, please check the script output on the host"
+    severity              = "Critical"
+    detect_label          = "UNKN"
+    disabled              = coalesce(var.status_check_disabled_critical, var.status_check_disabled, var.detectors_disabled)
```
xp-1000, line 15:
```suggestion
    disabled              = coalesce(var.status_check_disabled_unknown, var.status_check_disabled, var.detectors_disabled)
```

system/nagios-status-check/detectors-nagios-status-check.tf
```diff
@@ -0,0 +1,37 @@
...
+    notifications         = coalescelist(var.status_check_notifications_critical, var.status_check_notifications, var.notifications)
```
xp-1000, line 16:
```suggestion
    notifications         = coalescelist(var.status_check_notifications_unknown, var.status_check_notifications, var.notifications)
```

system/nagios-status-check/variables.tf
```diff
@@ -0,0 +1,87 @@
...
+  type        = bool
+  default     = null
+}
+
+variable "status_check_disabled_warning" {
+  description = "Disable warning alerting rule for status_check detector"
+  type        = bool
+  default     = null
+}
+
```
xp-1000, line 58: miss warning for disabled and notificaitons

system/nagios-status-check/variables.tf
```diff
@@ -0,0 +1,87 @@
...
+variable "status_check_aggregation_function" {
+  description = "Aggregation function and group by for status_check detector (i.e. \".mean(by=['host']).\")"
+  type        = string
+  default     = ""
+}
+
+variable "status_check_transformation_function" {
+  description = "Transformation function for status_check detector (mean, min, max)"
+  type        = string
+  default     = ".min(over='15m')"
```
xp-1000, line 86: what about using lasting instead of transofrmation function ?

cvauvarin: Changed of lasting.

system/nagios-status-check/variables.tf
```diff
@@ -0,0 +1,105 @@
...
+}
+
+variable "status_check_aggregation_function" {
+  description = "Aggregation function and group by for status_check detector (i.e. \".mean(by=['host']).\")"
+  type        = string
+  default     = ""
+}
+
+variable "status_check_transformation_function" {
+  description = "Transformation function for status_check detector (mean, min, max)"
```
xp-1000, line 96:
```suggestion
  description = ""Transformation function for status_check detector (i.e. \".mean(over='5m')\"))""
```

xp-1000 (approved):
LGTM

PR #36, opened 2020-07-24
use prometheus exporter for kong instead of collectd

xp-1000:
official kong monitor requires to install https://github.com/signalfx/kong-plugin-signalfx which is not the easiest way to do especially when the customer manage himself kong and claranet could not install it for him. Using promehteus is far easier even if it means we do not have the preset dashboard from signalfx official monitor https://docs.signalfx.com/en/latest/integrations/integrations-reference/integrations.kong.html#usage we could probably add a custom one in https://github.com/claranet/terraform-signalfx-dashboards/ in future

### DIFF
--- a/middleware/kong/README.md
+++ b/middleware/kong/README.md
@@ -53,4 +53,30 @@ Creates SignalFx detectors with the following checks:
 
 ## Related documentation
 
-[Official documentation](https://docs.signalfx.com/en/latest/integrations/integrations-reference/integrations.kong.html)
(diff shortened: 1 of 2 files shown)


xp-1000:
For information, as for Signalfx, Datadog also provides a plugin to get more metrics : 

- https://docs.konghq.com/hub/signalfx/signalfx/
- https://docs.konghq.com/hub/kong-inc/datadog/

which we do not use in my knowledge for the same reason of this PR.

PR #38, opened 2020-07-28
Unused variables

xp-1000:
crappy script to track any unused variable and fail the CI when meet


xp-1000:
see https://github.com/claranet/terraform-signalfx-detectors/runs/919618253?check_suite_focus=true for example

pdecat:
What about tflint? https://github.com/terraform-linters/tflint/pull/739

pdecat:
Or a python implementation https://alexwlchan.net/2019/05/finding-unused-variables-in-a-terraform-module/

xp-1000:
> Or a python implementation https://alexwlchan.net/2019/05/finding-unused-variables-in-a-terraform-module/

the pyhcl lib is not aims to fully support hcl, it fails on some of our modules (especially on locals and outputs) and implementation is long and do not allow exclusions.

> What about tflint? terraform-linters/tflint#739

better idea, I can test.

PR #76, opened 2020-09-01
use project_id instead of gcp_project_name which could change

xp-1000:


cloud/gcp/big-query/variables.tf
```diff
@@ -5,7 +5,7 @@ variable "environment" {
   type        = string
 }
 
-variable "gcp_project_name" {
+variable "gcp_project_id" {
   description = "GCP project name used for default filtering while lables are not synced"
```
pdecat, line 9:
```suggestion
  description = "GCP project id used for default filtering while labels are not synced"
```

pdecat (approved):

PR #81, opened 2020-09-07
fixing varnish memory space unit

nicolasvion:
- fixing varnish memory space unit (default is B which is 10^3)


xp-1000 (approved):
LGTM

xp-1000:
EDIT: space = available and bytes = allocated so need to sum them for a percentage

Issue #92, opened 2020-09-24
Generate outputs


xp-1000:
we should generate outputs for every detectors and probably add this to CI

xp-1000:
fixed by https://github.com/claranet/terraform-signalfx-detectors/pull/127

PR #127, opened 2020-09-29
generate outputs and check it in ci #92

xp-1000:


cloud/aws/alb/outputs.tf
```diff
@@ -0,0 +1,35 @@
+output "alb_4xx_id" {
+  description = "id for detector alb_4xx"
+  value       = signalfx_detector.alb_4xx.*.id
```
Shr3ps, line 3:
```suggestion
  value       = signalfx_detector.alb_4xx[*].id
```
cf https://www.terraform.io/docs/configuration/expressions.html#splat-expressions

Shr3ps (changes requested):

cloud/aws/alb/outputs.tf
xp-1000: While we do not use count like for datadog I will remove the list, thanks for the comment.

I updated PR to output the entire resource and not only the id.

Shr3ps (approved):

PR #131, opened 2020-10-05
Safer and easier notifications using rich types

xp-1000:
this PR includes:

- update varnish module to be compliant with usage practices (cc @nicolasvion )
- update nagios module to use existing signalfx severity for unknown alerts (cc @cvauvarin)
- replace global `notifications` variable to use object and force user to define every severities explicitely (safety feature to avoid undesired binding)
- replace per detector notifications variable to use simple map which allow use to only define relevant severities (not all) but will fail if required severity is not set
- delete per detector per severity variable as they now belongs to the per detector map variable

this should provide a safer but also easier approach to define notifications and remains flexible enough, every suggestions are welcome.

in breaking 0.12 compatibility we can probably go further using https://www.terraform.io/docs/configuration/variables.html#custom-validation-rules

### DIFF
--- a/cloud/aws/alb/detectors-alb.tf
+++ b/cloud/aws/alb/detectors-alb.tf
@@ -12,7 +12,7 @@ EOF
     severity              = "Critical"
     detect_label          = "CRIT"
     disabled              = coalesce(var.heartbeat_disabled, var.detectors_disabled)
-    notifications         = coalescelist(var.heartbeat_notifications, var.notifications)
+    notifications         = coalescelist(lookup(var.heartbeat_notifications, "critical", []), var.notifications.critical)
     parameterized_subject = "[{{ruleSeverity}}]{{{detectorName}}} {{{readableRule}}} on {{{dimensions}}}"
   }
 }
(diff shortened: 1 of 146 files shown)


cloud/aws/alb/detectors-alb.tf
```diff
@@ -12,7 +12,7 @@ EOF
     severity              = "Critical"
     detect_label          = "CRIT"
     disabled              = coalesce(var.heartbeat_disabled, var.detectors_disabled)
-    notifications         = coalescelist(var.heartbeat_notifications, var.notifications)
+    notifications         = coalescelist(lookup(var.heartbeat_notifications, "critical", []), var.notifications["critical"])
```
Shr3ps, line 15:
```suggestion
    notifications         = coalescelist(lookup(var.heartbeat_notifications, "critical", []), var.notifications.critical)
```

Shr3ps (changes requested):

jphilaine (approved):

BzSpi (approved):
:clap:

Shr3ps (approved):

cvauvarin (approved):

Issue #93, opened 2020-09-24
Rework severities level for translated detectors


xp-1000:
for now, the majority of detectors come from translation from datadog which only have 2 levels (alert, that we called critical and which is mandatory, and warning, a intermediate and optional threshold).

SignalFx provides 5 levels, all optional so we should take advantage of this changing in relevant way existing detectors.

Also we should document how we recommend to handle each severity (critical = oncall NBH, major = oncall BH only, minor = simple notification, warning = dashboard only ...) even if it remains only an advice and everybody could map each severity to recipient they want.

Also from the datadog legacy most of the detectors use "critical" and "warning" thresholds but "warning" threshold is not just below "critical" in signalfx convention while "major" and "minor" exist between them. We should probably follow the signalfx convention.

Some modules already use other severities like major and minor, especially those much reworked for translation (databases and kubernetes).

to do in same time https://github.com/claranet/terraform-signalfx-detectors/issues/94

PR #132, opened 2020-10-07
#93 reorder severities following signalfx levels

xp-1000:
replaces:
- warning by major
- major by minor
- minor by warning

the goal is to enjoy the multiple severity levels provided by signalfx compared to original datadog monitors.

we should probably decrease the severity of lot of legacy detectors because orginally we were forced to assign a critical/alert level for datadog but now we can define any level on signalfx.

this could be done little by little in future, this PR only aims to reorder severities and keep them as they are.

### DIFF
--- a/cloud/aws/alb/variables.tf
+++ b/cloud/aws/alb/variables.tf
@@ -82,8 +82,8 @@ variable "no_healthy_instances_disabled_critical" {
   default     = null
 }
 
-variable "no_healthy_instances_disabled_warning" {
-  description = "Disable warning alerting rule for No_healthy_instances detector"
+variable "no_healthy_instances_disabled_major" {
+  description = "Disable major alerting rule for No_healthy_instances detector"
   type        = bool
   default     = null
 }
@@ -112,8 +112,8 @@ variable "no_healthy_instances_threshold_critical" {
   default     = 1
 }
 
-variable "no_healthy_instances_threshold_warning" {
-  description = "Warning threshold for No_healthy_instances detector"
+variable "no_healthy_instances_threshold_major" {
+  description = "Major threshold for No_healthy_instances detector"
   type        = number
   default     = 100
 }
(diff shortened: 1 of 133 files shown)


Shr3ps (approved):

cloud/aws/alb/variables.tf
```diff
@@ -11,9 +11,9 @@ variable "notifications" {
   description = "Default notification recipients list per severity"
   type = object({
     critical = list(string)
-    major    = list(string)
```
jphilaine, line 14: The order of severities doesn't respect the SignalFX order.
It's better to have :
- critical
- major
- minor
- warning
- info

This change appears into all modules.

cloud/azure/app-service/variables.tf
```diff
@@ -106,7 +106,7 @@ variable "response_time_threshold_critical" {
   default     = 10
 }
 
-variable "response_time_threshold_warning" {
+variable "response_time_threshold_major" {
   description = "Warning threshold for response_time detector"
```
jphilaine, line 110: It's not Warning threshold, but Major. You need replace all `Warning` by `Major`

jphilaine (approved):

cvauvarin:
To be tested but I am quite sure that every alert received by relkon will end in the `critical` severity by default. This is not only a problem with the dashboard as the right severity will not be displayed but it will lead to unwanted alert during non business hour.

cvauvarin:
To confirm this, an extract of the pipeline in relkon : 

The status field in relkon is mapped to the severity from SignalFX.

```
    if [status] =~ /(OK|RECOVERY|recovery|Recovery|RECOVERED|recovered|Recovered)/ {
        mutate {
            replace => { "status" => "OK" }
            replace => { "event" => "RECOVERY" }
        }
     }
    else if [status] =~ /(WARNING|warning|Warning|Warn)/ {
                mutate {
                        replace => { "status" => "WARNING" }
        }
    }
    else if [status] =~ /(UNKNOWN|unknown|Unknown)/ {
        mutate {
            replace => { "status" => "UNKNOWN" }
        }
    }
    else {
        mutate {
            replace => { "status" => "CRITICAL" }
        }
    }
```

There are two solutions :
* patch this pipeline to map the status field to a valid one
* do the mapping in the relkon API directly

BzSpi (approved):

cvauvarin (approved):

PR #143, opened 2020-10-14
[Azure] PostGreSQL/MySQL detectors - rework storage detectors

Shr3ps:
detector now based on "usage" instead of "free space"


xp-1000 (approved):
good

PR #150, opened 2020-10-16
#146 network/http add lasting for code check

### LINKED ISSUE #146: network/http - Allow to make http detectors less sensitive
On http webchecks, alerts triggers too fast and creates false positives.

I think we should use lasting() + variables to customize the duration before triggering an alert.

jmapro:
Allow to retard alerts by using lasting(). Default value to 1s to not break existing code.

Fix #146

### DIFF
--- a/network/http/detectors-http.tf
+++ b/network/http/detectors-http.tf
@@ -24,7 +24,7 @@ resource "signalfx_detector" "http_code_matched" {
 
   program_text = <<-EOF
     signal = data('http.code_matched', ${module.filter-tags.filter_custom})${var.http_code_matched_aggregation_function}${var.http_code_matched_transformation_function}.publish('signal')
-    detect(when(signal < 1)).publish('CRIT')
+    detect(when(signal < 1, lasting='${var.http_code_matched_lasting_duration_seconds}s', at_least=${var.http_code_matched_at_least_percentage})).publish('CRIT')
 EOF
 
   rule {
@@ -44,7 +44,7 @@ resource "signalfx_detector" "http_regex_matched" {
 
   program_text = <<-EOF
     signal = data('http.regex_matched', ${module.filter-tags.filter_custom})${var.http_regex_matched_aggregation_function}${var.http_regex_matched_transformation_function}.publish('signal')
-    detect(when(signal < 1)).publish('CRIT')
+    detect(when(signal < 1, lasting='${var.http_regex_matched_lasting_duration_seconds}s', at_least=${var.http_regex_matched_at_least_percentage})).publish('CRIT')
 EOF
 
   rule {
--- a/network/http/variables.tf
+++ b/network/http/variables.tf
@@ -169,7 +169,7 @@ variable "http_response_time_threshold_major" {
 variable "http_content_length_disabled" {
   description = "Disable all alerting rules for http_content_length detector"
   type        = bool
-  default     = null
+  default     = true
 }
 
 variable "http_content_length_notifications" {
--- a/network/http/variables.tf
+++ b/network/http/variables.tf
@@ -55,7 +55,19 @@ variable "http_code_matched_aggregation_function" {
 variable "http_code_matched_transformation_function" {
   description = "Transformation function for http_code_matched detector (i.e. \".mean(over='5m')\")"
   type        = string
-  default     = ".max(over='1m')"
+  default     = ""
+}
+
+variable "http_code_matched_lasting_duration_seconds" {
+  description = "Duration that indicates how long we wait before triggering the http_code_matched alert. Specified as s, m, h, d"
+  type        = number
+  default     = 60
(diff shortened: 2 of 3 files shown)


network/http/detectors-http.tf
```diff
@@ -24,7 +24,7 @@ resource "signalfx_detector" "http_code_matched" {
 
   program_text = <<-EOF
     signal = data('http.code_matched', ${module.filter-tags.filter_custom})${var.http_code_matched_aggregation_function}${var.http_code_matched_transformation_function}.publish('signal')
-    detect(when(signal < 1)).publish('CRIT')
+    detect(when(signal < 1, lasting('${var.http_code_matched_lasting_duration}'))).publish('CRIT')
```
xp-1000, line 27: please follow existing implementation which force seconds as unit to make it possible arithmetical operations inside terraform: https://github.com/claranet/terraform-signalfx-detectors/blob/master/cloud/aws/alb/detectors-alb.tf#L58

```suggestion
    detect(when(signal < 1, lasting(asting='${var.http_code_matched_lasting_duration_seconds}s'))).publish('CRIT')
```

network/http/detectors-http.tf
```diff
@@ -44,7 +44,7 @@ resource "signalfx_detector" "http_regex_matched" {
 
   program_text = <<-EOF
     signal = data('http.regex_matched', ${module.filter-tags.filter_custom})${var.http_regex_matched_aggregation_function}${var.http_regex_matched_transformation_function}.publish('signal')
-    detect(when(signal < 1)).publish('CRIT')
+    detect(when(signal < 1, lasting('${var.http_regex_matched_lasting_duration}'))).publish('CRIT')
```
xp-1000, line 47: please follow existing implementation which force seconds as unit to make it possible arithmetical operations inside terraform: https://github.com/claranet/terraform-signalfx-detectors/blob/master/cloud/aws/alb/detectors-alb.tf#L58

```suggestion
     detect(when(signal < 1, lasting(asting='${var.http_regex_matched_lasting_duration_seconds}s'))).publish('CRIT')
```

network/http/variables.tf
```diff
@@ -58,6 +58,12 @@ variable "http_code_matched_transformation_function" {
   default     = ".max(over='1m')"
 }
 
+variable "http_code_matched_lasting_duration" {
+  description = "Duration that indicates how long we wait before triggering the http_code_matched alert. Specified as s, m, h, d"
+  type        = string
+  default     = "1s"
+}
+
```
xp-1000, line 66: please follow current naming: https://github.com/claranet/terraform-signalfx-detectors/blob/master/cloud/aws/alb/variables.tf#L123

```suggestion
variable "http_code_matched_lasting_duration_seconds" {
  description = "Minimum duration that conditions must be true before raising alert (in seconds)"
  type        = number
  default     = 60
}

```

I don't understand why you choose `1s` which will not make this detector more tolerant (which is what you want to achieve I think) so I suggested `60s` but feel free to change as you prefer.

network/http/variables.tf
```diff
@@ -90,6 +96,12 @@ variable "http_regex_matched_transformation_function" {
   default     = ".max(over='5m')"
 }
 
+variable "http_regex_matched_lasting_duration" {
+  description = "Duration that indicates how long we wait before triggering the http_regex_matched alert. Specified as s, m, h, d"
+  type        = string
+  default     = "1s"
+}
+
```
xp-1000, line 104: please follow current naming: https://github.com/claranet/terraform-signalfx-detectors/blob/master/cloud/aws/alb/variables.tf#L123

```suggestion
variable "http_regex_matched_lasting_duration_seconds" {
  description = "Minimum duration that conditions must be true before raising alert (in seconds)"
  type        = number
  default     = 60
}

```

I don't understand why you choose `1s` which will not make this detector more tolerant (which is what you want to achieve I think) so I suggested `60s` but feel free to change as you prefer.

xp-1000 (changes requested):
some remarks:

- I prefer to follow existing convention (naming, implementation of lasting)
- keep in mind this is already possible to make this detector more tolerant increasing the transformation function timeframe
- keep in mind it is possible to also configure the `at least` option of lasting function
- adding a lasting at "1s" seems useless and need more (in my opinion, at least 20s while the default interval is 10s)

xp-1000:
just seen your message

> Default value to 1s to not break existing code.

the monitor is still new and developped especially by Claranet so I think you an go ahead and change the default behavior as you prefer while you are one of the first user.

Nevertheless, no matter you prefer to keep the current "strict" implementation or change for a "tolerant" one you should explain this choice in the readme because this is opinionated and if default behavior should please to the majority of users there is always few ones who will want the strict behavior

xp-1000 (approved):
Good thanks!

Issue #123, opened 2020-09-29
Cloud Azure - Capacity detectors


Shr3ps:
Review or add detectors on each Azure resource to be able to monitor resource capacity:
- instance usage limit
- scaling limit
- sku/type
- RUs, DTUs, perfs limit

Claranet ref [AZ-270]

PR #163, opened 2020-10-26
#123 [Azure] CosmosDB capacity (RUs usage) detector

Shr3ps:


xp-1000 (approved):
I am not a big fan of using uppercase (RU) in variables names but it looks good thanks.

Shr3ps:
> 
> 
> I am not a big fan of using uppercase (RU) in variables names but it looks good thanks.

I will plan to make some deeper tests by the end of the week, and yes I will lower variables (I kept what was initially generated by the template j2 ;) )

xp-1000:
> > I am not a big fan of using uppercase (RU) in variables names but it looks good thanks.
> 
> I will plan to make some deeper tests by the end of the week, and yes I will lower variables (I kept what was initially generated by the template j2 ;) )

good input I will take care of that.

PR #170, opened 2020-10-30
memcached: bootstrap detectors

cvauvarin:


xp-1000 (approved):
seems good. some suggestions for other detectors:
- hit ratio detector (implemented on aws memcached) eithier using `memcached_ops.hits / memcached_command.get * 100` or `memcached_ops.hits / (memcached_ops.hits + memcached_ops.misses) * 100` or directly using `percent.hitratio` metrics which should be available if added to `extraMetrics`.
- fill percent detector using `df.cache.free / (df.cache.free + df.cache.used) * 100`.
- having `memcached_ops.evictions > 0` is not a problem itself but if "fill percent" is < 100% so this a single slab class is probably full

PR #174, opened 2020-11-03
#120 add mongo detectors

### LINKED ISSUE #120: Detectors for mongodb
collectd monitor seems not enough https://docs.signalfx.com/en/latest/integrations/agent/monitors/collectd-mongodb.html 

for example it miss replication metrics. May be the best way is to create a new monitor from telegraf plugin https://github.com/influxdata/telegraf/tree/master/plugins/inputs/mongodb

and use it to create detectors

xp-1000:
fix #120

### DIFF
--- a/modules/database-mongo/README.md
+++ b/modules/database-mongo/README.md
@@ -28,10 +28,20 @@ This module uses [collectd/mongodb](https://docs.signalfx.com/en/latest/integrat
 
 ## Notes
 
-* Primary and secondary detectors require to configure on all members 
-of the replicat because they use explicitly aggregation by replicaset 
-(`cluster` by default) to work. Change default value of corresponding 
-`aggregation_function` variable if necessary.
+* This is highly recommended to configure the `collectd/mongodb` 
+monitor for __all__ members of a replicaset to fetch metrics from 
+both the primary and the secondaries which all have their own stats.
 
-* The heartbeat detector is by aggregated replicaset (`cluster`) by 
+* This is mandatory for "primary" and "secondary" detectors because 
+they need the information of every member of a replicaset to determine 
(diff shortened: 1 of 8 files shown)


markyam (approved):

modules/database-mongo/README.md
```diff
@@ -0,0 +1,37 @@
...
+```
+
+## Notes
+
+* Primary and secondary detectors require to configure on all members 
+of the replicat because they use explicitly aggregation by replicaset 
+(`cluster` by default) to work. Change default value of corresponding 
+`aggregation_function` variable if necessary.
+
+* The heartbeat detector is by aggregated replicaset (`cluster`) by
```
cvauvarin, line 36: is aggregated by replicaset

modules/database-mongo/README.md
```diff
@@ -0,0 +1,37 @@
...
+      - counter.asserts.warning
+      # Only required if agent <= 5.5.5:
+      - gauge.repl.max_lag
+      - gauge.repl.active_nodes
+      - gauge.repl.is_primary_node
+```
+
+## Notes
+
+* Primary and secondary detectors require to configure on all members
```
cvauvarin, line 31: I do not understand what you mean by "require to configure on all members". I think you need to rephrase the statement.

modules/database-mongo/detectors-mongo.tf
```diff
@@ -0,0 +1,163 @@
+resource "signalfx_detector" "heartbeat" {
+  name      = format("%s %s", local.detector_name_prefix, "Webcheck heartbeat")
```
cvauvarin, line 2: Webcheck -> Mongodb

cvauvarin (approved):

Issue #156, opened 2020-10-22
Helper to create new detectors


xp-1000:
Waiting for https://github.com/claranet/terraform-signalfx-detectors/issues/105 or even in complementary it could be nice to create a generator to help writing new detectors as terraform code and ease contributions.

PR #175, opened 2020-11-03
#156 various improvements for generator

xp-1000:
improve #156 preserving terraform fmt for variables (trim ending whitespace) and avoid capitalize which lower every chars except the first one (MongoDB -> Mongodb)

EDIT: adding threshold detectors to ci tests showed another problem with `name` format when `max_delay` is set (for heartbeat only).

- trim "for" based starting variables declaration:
(`# xyz detector` -> `# xyz detector`)

- do not lower module name (but keep upper for first char)

- terraform format for `name` when `max_delay` is defined (heartbeat)

- fix hardcoded comparator in main detect condition

- fix threshold to 0 is considered as not defined

- ability to override id for long name or containing special chars

- add base filtering support

### DIFF
--- a/scripts/templates/detector_vars.tf.j2
+++ b/scripts/templates/detector_vars.tf.j2
@@ -1,4 +1,8 @@
-{%- set id = name | replace(' ', '_') -%}
+{%- if id is defined and id -%}
+  {%- set id = id | lower -%}
+{%- else -%}
+  {%- set id = name | replace(' ', '_') | lower -%}
+{% endif -%}
 {%- if name | lower == 'heartbeat' -%}
   {%- set type = 'heartbeat' -%}
 {%- else -%}
@@ -47,7 +51,7 @@ variable "{{ id }}_disabled_{{ severity }}" {
   default     = null
 }
 
-  {% endfor -%}
+{% endfor -%}
 {% endif -%}
 
 {% if type == 'heartbeat' -%}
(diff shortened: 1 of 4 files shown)


scripts/templates/detector_vars.tf.j2
```diff
@@ -47,7 +47,7 @@ variable "{{ id }}_disabled_{{ severity }}" {
   default     = null
 }
 
-  {% endfor -%}
+{% endfor -%}
```
Shr3ps, line 50: why did you drop this indent?

xp-1000: to avoid adding whitespace at the end and keep the file "terraform fmt"

Shr3ps: what about using `{%- endfor -%}` for this instead (the - symbol should ignore indent)

xp-1000: because if it will ignore the indent for sure it will also remove previous blank line which is desired.

feel free to try, I think it is possible but not without to change the "logic" of the entire "for" block.
Keep in mind it should produce a "terraform fmt-ed" output and consistent in every scenarios (with aggregation / transformation and without them).

while this indent is only here for readability purpose I confess I did not look so far and used the easy way and prefer to avoid too much changes.

the real problem is the lack of testing. adding more test cases in "generator ci" will allow us to trust our changes.

xp-1000: ok I added every threshold detectors to tests for generator.

here is the result before this PR changes: https://github.com/claranet/terraform-signalfx-detectors/runs/1349218290?check_suite_focus=true

you can see the problems caused by this endfor indent like:

```
-  variable "requests_errors_5xx_percentage_threshold_major" {
+variable "requests_errors_5xx_percentage_threshold_major" {
```

or 

```
-  resource "signalfx_detector" "real_time_requests_latency" {
+resource "signalfx_detector" "real_time_requests_latency" {
```

now there is the result with this PR changes applied: https://github.com/claranet/terraform-signalfx-detectors/runs/1349338707?check_suite_focus=true

now we could be more confident about submit bigger change. 
That said, I still think this indent is not essential and its deletion is good enough to keep the current logic (which seems good enough too).

Shr3ps (approved):

PR #178, opened 2020-11-09
improve http detectors:

xp-1000:
- handle "duplicated" monitor config using extrapolation
- better severities definition
- cosmetic for alert rule name

### DIFF
--- a/network/http/detectors-http.tf
+++ b/network/http/detectors-http.tf
@@ -20,15 +20,15 @@ EOF
 }
 
 resource "signalfx_detector" "http_code_matched" {
-  name = format("%s %s", local.detector_name_prefix, "HTTP code ")
+  name = format("%s %s", local.detector_name_prefix, "HTTP code")
 
   program_text = <<-EOF
-    signal = data('http.code_matched', filter=${module.filter-tags.filter_custom})${var.http_code_matched_aggregation_function}${var.http_code_matched_transformation_function}.publish('signal')
+    signal = data('http.code_matched', filter=${module.filter-tags.filter_custom}, rollup='max')${var.http_code_matched_aggregation_function}${var.http_code_matched_transformation_function}.publish('signal')
     detect(when(signal < 1, lasting='${var.http_code_matched_lasting_duration_seconds}s', at_least=${var.http_code_matched_at_least_percentage})).publish('CRIT')
 EOF
 
   rule {
-    description           = "does not match"
+    description           = "does not match expected result"
     severity              = "Critical"
     detect_label          = "CRIT"
     disabled              = coalesce(var.http_code_matched_disabled_critical, var.http_code_matched_disabled, var.detectors_disabled)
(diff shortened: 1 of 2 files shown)


Gati0 (approved):

network/http/detectors-http.tf
```diff
@@ -20,15 +20,15 @@ EOF
...
 
   program_text = <<-EOF
-    signal = data('http.code_matched', filter=${module.filter-tags.filter_custom})${var.http_code_matched_aggregation_function}${var.http_code_matched_transformation_function}.publish('signal')
+    signal = data('http.code_matched', filter=${module.filter-tags.filter_custom}, rollup='max')${var.http_code_matched_aggregation_function}${var.http_code_matched_transformation_function}.publish('signal')
     detect(when(signal < 1, lasting='${var.http_code_matched_lasting_duration_seconds}s', at_least=${var.http_code_matched_at_least_percentage})).publish('CRIT')
 EOF
 
   rule {
-    description           = "does not match"
+    description           = "does not match expected"
```
jmapro, line 31: We may put "does not match expected HTTP code" ?

network/http/detectors-http.tf
```diff
@@ -40,15 +40,15 @@ EOF
...
 
   program_text = <<-EOF
-    signal = data('http.regex_matched', filter=${module.filter-tags.filter_custom})${var.http_regex_matched_aggregation_function}${var.http_regex_matched_transformation_function}.publish('signal')
+    signal = data('http.regex_matched', filter=${module.filter-tags.filter_custom}, rollup='min')${var.http_regex_matched_aggregation_function}${var.http_regex_matched_transformation_function}.publish('signal')
     detect(when(signal < 1, lasting='${var.http_regex_matched_lasting_duration_seconds}s', at_least=${var.http_regex_matched_at_least_percentage})).publish('CRIT')
 EOF
 
   rule {
-    description           = "does not match"
+    description           = "does not match in body response"
```
jmapro, line 51: I'd rather write "Expression not found in body response"

network/http/detectors-http.tf
```diff
@@ -93,47 +93,47 @@ resource "signalfx_detector" "http_content_length" {
...
     parameterized_body    = local.rule_body
   }
 
   rule {
-    description           = " < ${var.certificate_expiration_date_threshold_major} days"
-    severity              = "Major"
-    detect_label          = "MAJOR"
-    disabled              = coalesce(var.certificate_expiration_date_disabled_major, var.certificate_expiration_date_disabled, var.detectors_disabled)
-    notifications         = coalescelist(lookup(var.certificate_expiration_date_notifications, "major", []), var.notifications.major)
+    description           = "is too low < ${var.certificate_expiration_date_threshold_minor} days"
```
jmapro, line 132: `     description           = "expires very soon: ${var.certificate_expiration_date_threshold_minor} days"`

network/http/detectors-http.tf
```diff
@@ -143,18 +143,17 @@ resource "signalfx_detector" "invalid_tls_certificate" {
...
 
   program_text = <<-EOF
-    signal = data('http.cert_valid', filter=${module.filter-tags.filter_custom})${var.invalid_tls_certificate_aggregation_function}${var.invalid_tls_certificate_transformation_function}.publish('signal')
+    signal = data('http.cert_valid', filter=${module.filter-tags.filter_custom}, rollup='min')${var.invalid_tls_certificate_aggregation_function}${var.invalid_tls_certificate_transformation_function}.publish('signal')
     detect(when(signal < 1)).publish('CRIT')
 EOF
 
   rule {
-    description           = " is not valid"
+    description           = "is not valid"
```
jmapro, line 151: `description           = "is invalid"`

BzSpi:
Can you also add units on alert messages when pertinent ? I mostly think of adding "second" in "response time" message (I cannot comment the affected line).

network/http/detectors-http.tf
BzSpi: The alert message is concatenated with the detector name. Currently, the full message will be "HTTP code does not match expected"

xp-1000: so we can put "does not match the expecting one" if you prefer ^^

network/http/detectors-http.tf
xp-1000: If I do that I will have to remove "expiring" from the name and it will be equal to the invalid detector one. Name are not used as ID and this is technically possible but I prefer to keep every name unique to list them in readme for example.

I propose to set : 

- the name as "TLS certificate expiry date"
- rule as "is too soon < X days"

what do you think?

jmapro: Seems good to me

jmapro (approved):

BzSpi (approved):

Issue #97, opened 2020-09-24
Module readmes


xp-1000:
we should have a readme for each module with minimum information to use them, especially every specificities compared to other modules.

obviously we need to make them more consistent with a template to follow (and free field).

should be probably generated and added in CI.

PR #182, opened 2020-11-20
#97 rework arbo and add script to gen modules readmes

xp-1000:
Refactor arbo into separate:

- module
- stack

files like scripts or common dirs to make it clear if it targets final users (stack) or contributors (module).

a new script to bootstrap stack is also added but a better refactor on all scripts will be done in another PR

please do not be too critic about the `gen_doc.sh` script while it will be simplified and cleaned after moving all modules to `/modules` directory.

so be focus on the jinja template and eventually the CI changes, for now the script implementation is temporary.

also added an example of new readme for the nginx module, see this PR https://github.com/claranet/terraform-signalfx-detectors/pull/183 for the full and final readmes.


middleware/nginx/README.md
```diff
@@ -1,55 +1,83 @@
...
+
+* Nginx dropped connections
+* Nginx heartbeat
+
+## How to collect required metrics?
+
+This module uses metrics collected from the [SignalFx Smart 
+Agent](https://github.com/signalfx/signalfx-agent).
+
+
```
Shr3ps, line 73: Lots of new line here 😝

Shr3ps (approved):

middleware/nginx/README.md
xp-1000: not a problem for markdown

scripts/templates/readme.md.j2
```diff
@@ -0,0 +1,108 @@
+# {{ module | upper }} SignalFx detectors
+
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## How do you use this module?
```
cvauvarin, line 8: `How to use this module` is better I think.

scripts/templates/readme.md.j2
```diff
@@ -0,0 +1,108 @@
...
+The [guidance](https://github.com/claranet/terraform-signalfx-detectors/wiki/Guidance) documentation will help you to use 
+common mechanims provided by the modules like [multi 
+instances](https://github.com/claranet/terraform-signalfx-detectors/wiki/Guidance#Multiple-instances).
+
+Feel free to explore the [wiki](https://github.com/claranet/terraform-signalfx-detectors/wiki) for more information about 
+general usage of this repository.
+
+## What are the available detectors in this module?
+
+This module creates the following SignalFx detectors which could contain one or mulitple alerting rules:
```
cvauvarin, line 58: `mulitple` -> `multiple`

cvauvarin (approved):

PR #185, opened 2020-11-24
[Azure][MySQL] Add failed_connections detector

markyam:
HI,

This PR is for a new detector on MySQL Azure to check the failed connections.

### DIFF
--- a/cloud/azure/mysql/detectors-mysql.tf
+++ b/cloud/azure/mysql/detectors-mysql.tf
@@ -175,7 +175,7 @@ resource "signalfx_detector" "replication_lag" {
 }
 
 resource "signalfx_detector" "failed_connections" {
-  name = format("%s %s", local.detector_name_prefix, "Mysql failed connections")
+  name = format("%s %s", local.detector_name_prefix, "Azure MySQL failed connections")
 
   program_text = <<-EOF
     base_filter = filter('resource_type', 'Microsoft.DBforMySQL/servers') and filter('primary_aggregation_type', 'true')
--- a/cloud/azure/mysql/detectors-mysql.tf
+++ b/cloud/azure/mysql/detectors-mysql.tf
@@ -178,21 +178,11 @@ resource "signalfx_detector" "failed_connections" {
   name = format("%s %s", local.detector_name_prefix, "Mysql failed connections")
 
   program_text = <<-EOF
-    base_filtering = filter('primary_aggregation_type', 'true')
-    signal = data('connections_failed', filter=base_filtering and ${module.filter-tags.filter_custom})${var.failed_connections_aggregation_function}${var.failed_connections_transformation_function}.publish('signal')
-    detect(when(signal > ${var.failed_connections_threshold_major})).publish('MAJOR')
+    base_filter = filter('resource_type', 'Microsoft.DBforMySQL/servers') and filter('primary_aggregation_type', 'true')
+    signal = data('connections_failed', filter=base_filter and ${module.filter-tags.filter_custom})${var.failed_connections_aggregation_function}${var.failed_connections_transformation_function}.publish('signal')
     detect(when(signal > ${var.failed_connections_threshold_critical})).publish('CRIT')
+    detect(when(signal > ${var.failed_connections_threshold_major})).publish('MAJOR')
 EOF
-
-  rule {
-    description           = "is too high > ${var.failed_connections_threshold_major}"
-    severity              = "Major"
-    detect_label          = "MAJOR"
-    disabled              = coalesce(var.failed_connections_disabled_major, var.failed_connections_disabled, var.detectors_disabled)
-    notifications         = coalescelist(lookup(var.failed_connections_notifications, "major", []), var.notifications.major)
-    parameterized_subject = local.rule_subject
-    parameterized_body    = local.rule_body
-  }
 
   rule {
     description           = "is too high > ${var.failed_connections_threshold_critical}"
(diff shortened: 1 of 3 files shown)


cloud/azure/mysql/detectors-mysql.tf
```diff
@@ -173,3 +173,34 @@ resource "signalfx_detector" "replication_lag" {
     parameterized_body    = local.rule_body
   }
 }
+
+resource "signalfx_detector" "failed_connections" {
+  name = format("%s %s", local.detector_name_prefix, "Mysql failed connections")
```
Shr3ps, line 178:
```suggestion
  name = format("%s %s", local.detector_name_prefix, "Azure MySQL failed connections")
```

cloud/azure/mysql/variables.tf
```diff
@@ -275,3 +275,53 @@ variable "replication_lag_threshold_major" {
...
+variable "failed_connections_notifications" {
+  description = "Notification recipients list per severity overridden for failed_connections detector"
+  type        = map(list(string))
+  default     = {}
+}
+
+variable "failed_connections_aggregation_function" {
+  description = "Aggregation function and group by for failed_connections detector (i.e. \".mean(by=['host'])\")"
+  type        = string
+  default     = ""
```
Shr3ps, line 290:
```suggestion
  default     = ".mean(by=['azure_resource_name', 'azure_resource_group_name', 'azure_region'])"
```

cloud/azure/mysql/variables.tf
```diff
@@ -275,3 +275,53 @@ variable "replication_lag_threshold_major" {
...
+}
+
+variable "failed_connections_transformation_function" {
+  description = "Transformation function for failed_connections detector (i.e. \".mean(over='5m')\")"
+  type        = string
+  default     = ""
```
Shr3ps, line 296: no default transformation?

cloud/azure/mysql/variables.tf
```diff
@@ -275,3 +275,53 @@ variable "replication_lag_threshold_major" {
...
+variable "failed_connections_threshold_major" {
+  description = "Major threshold for failed_connections detector"
+  type        = number
+  default     = 10
+}
+
+variable "failed_connections_threshold_critical" {
+  description = "Critical threshold for failed_connections detector"
+  type        = number
+  default     = 20
```
Shr3ps, line 326: not sure about this value, maybe it may be better to compare with `active_connections` metric?

Shr3ps (changes requested):

cloud/azure/mysql/detectors-mysql.tf
```diff
@@ -173,3 +173,34 @@ resource "signalfx_detector" "replication_lag" {
...
 }
+
+resource "signalfx_detector" "failed_connections" {
+  name = format("%s %s", local.detector_name_prefix, "Azure MySQL failed connections")
+
+  program_text = <<-EOF
+    base_filter = filter('resource_type', 'Microsoft.DBforMySQL/servers') and filter('primary_aggregation_type', 'true')
+    signal = data('connections_failed', filter=base_filter and ${module.filter-tags.filter_custom})${var.failed_connections_aggregation_function}${var.failed_connections_transformation_function}.publish('signal')
+    detect(when(signal > ${var.failed_connections_threshold_critical})).publish('CRIT')
+    detect(when(signal > ${var.failed_connections_threshold_major})).publish('MAJOR')
```
xp-1000, line 184: it miss dependency with critical to avoid double notif

cloud/azure/mysql/variables.tf
xp-1000: it depends on what this metric is and sadly the azure doc is not very useful as usual: https://docs.microsoft.com/en-us/azure/azure-monitor/platform/metrics-supported#microsoftdbformysqlservers

in my knowledge of mysql monitoring world I only know two metrics (from mysql point of view) which could match:
- [aborted clients](https://dev.mysql.com/doc/refman/5.6/en/server-status-variables.html#statvar_Aborted_clients)
- [aborted connects](https://dev.mysql.com/doc/refman/5.6/en/server-status-variables.html#statvar_Aborted_connects)

I doubt Microsoft collects more advanced data to create this metric but it could be a sum of both :man_shrugging: 

assuming no and it simply is the popular `aborted_connects` value so I agree with @Shr3ps it seems not relevant to alert on a static threshold here considering the number of "normal" errors it could happen: https://dev.mysql.com/doc/refman/5.6/en/communication-errors.html. 

I suppose wrong password could be a security problem of course but for now our modules are not security focused and if we add security oriented alerts I think it should be in separate modules than the "basic" infra alerts purpose.

on the [agent based version of this module](https://github.com/claranet/terraform-signalfx-detectors/tree/master/database/mysql) I provide a way in readme to collect this metric but I don't use it in detectors because I think the only relevant error to alert is the max_conections reached scenario which is better monitored "before" the error happen.

Here you don't have the max_connections value and as I understand https://docs.microsoft.com/en-us/azure/mysql/concepts-limits#server-parameters it could be tuned so we cannot trust the instance type or default value config so it seems impossible to calculate the % of connections over max capacity.

Without that it makes sens to use this metric even if it could contains not relevant errors.
Using the active connection as suggested by @Shr3ps could mitigate this risk but I don't think make a simple percentage is a good idea. it will probably never raise alert, too late or false.

Of course with more advanced work like double "AND" conditions to correlate:
- the number of failed connections value > 0
- with a "% of change" of active connections number ~=0% (i.e. `.rateofchange().abs()`)
we can suppose if errors number increases in the same time the active connections number stagnate this is probably a max connection reached scenario.

may be third condition to compare the value of active connections from the present to a little `timeshift` like 15m will help to be sure it is a valid alert but it could lead to auto resolve after a long period

Else I only see workarounds:
- to simply force the user to set the threshold depending on the max connections limit and make a detector (in separate module than this one because will be used "per" database). in this case you alert on max connection reached and you do not need this metric anymore.
- to use https://docs.signalfx.com/en/latest/detect-alert/alert-condition-reference/hist-anomaly.html instead of threshold based detector on failed or even number of active connections. it will not really alert on max connections reached (as for your current detector) but should be smarter and avoid "stupid" alerts.

xp-1000 (changes requested):
For now I don't think this detector is judicious.

you should either:
- consider to give up with cloud providers metrics and focus on agent based collection and alerting
- change of direction (see workarounds proposed)
- enrich the signalflow programm to mitigate the high risk of not relevant alerts of the current implementation

markyam:
You can close this PR, we decided to use the aborted_connect metric.

PR #187, opened 2020-11-25
Use regex instead of replace to only split on first `:` in filter

### LINKED ISSUE #186: filter-tags variables should only split first ":"
as per https://github.com/claranet/terraform-signalfx-detectors/blob/master/common/filter-tags/locals.tf#L14 , filter-tags variables replace every instance of `:` with `', '`. This breaks filters with more than one `:`, for ex in http detector (with filters like `url:http://github.com`

maxenced:
fix #186 

The use of `regex` to match only first `:` avoid to split extra part of string.

### DIFF
--- a/common/filter-tags/locals.tf
+++ b/common/filter-tags/locals.tf
@@ -3,7 +3,7 @@ locals {
     " and ",
     formatlist(
       "(filter('%s'))",
-      [for tag in var.filter_custom_includes : replace(replace(tag, ":", "', '"), " ", "")]
+      [for tag in var.filter_custom_includes : replace(join("', '", regex("^([^:]+):(.*)", tag)), " ", "")]
     )
   )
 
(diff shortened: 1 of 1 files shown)


common/filter-tags/locals.tf
```diff
@@ -3,15 +3,15 @@ locals {
...
+      [for tag in var.filter_custom_includes : replace(join("', '", regex("^([^:]+):(.*)", tag)), " ", "")]
     )
   )
 
   filter_custom_excludes = join(
     " and ",
     formatlist(
       "(not filter('%s'))",
-      [for tag in var.filter_custom_excludes : replace(replace(tag, ":", "', '"), " ", "")]
+      [for tag in var.filter_custom_excludes : replace(join("', '", regex("^([^:]+):(.*)", k)), " ", "")]
```
xp-1000, line 14:
```suggestion
      [for tag in var.filter_custom_excludes : replace(join("', '", regex("^([^:]+):(.*)", tag)), " ", "")]
```

xp-1000 (changes requested):
it seems you miss something in the excludes version

xp-1000 (approved):
It is good for me thanks for that!

PR #189, opened 2020-11-26
haproxy: do not alert on backend limit by default

cvauvarin:
There are no real limits on the backend in Haproxy but there are on the `server` and `frontend` components.

The limit parameter on the `backend` is useful only if there is a `minconn` parameter on the server (see https://www.mail-archive.com/haproxy@formilux.org/msg26566.html). 

I think it is better to filter only on `server` and `frontend` by default to avoid unnecessary alerts.

### DIFF
--- a/middleware/haproxy/detectors-haproxy.tf
+++ b/middleware/haproxy/detectors-haproxy.tf
@@ -61,7 +61,7 @@ resource "signalfx_detector" "session_limit" {
   name = format("%s %s", local.detector_name_prefix, "Haproxy session")
 
   program_text = <<-EOF
-    A = data('haproxy_session_current', filter=${module.filter-tags.filter_custom})${var.session_limit_aggregation_function}${var.session_limit_transformation_function}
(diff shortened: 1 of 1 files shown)


middleware/haproxy/detectors-haproxy.tf
```diff
@@ -61,7 +61,7 @@ resource "signalfx_detector" "session_limit" {
   name = format("%s %s", local.detector_name_prefix, "Haproxy session")
 
   program_text = <<-EOF
-    A = data('haproxy_session_current', filter=${module.filter-tags.filter_custom})${var.session_limit_aggregation_function}${var.session_limit_transformation_function}
+    A = data('haproxy_session_current', filter=${var.session_limit_filter} and ${module.filter-tags.filter_custom})${var.session_limit_aggregation_function}${var.session_limit_transformation_function}
```
xp-1000, line 64: why not hardcode directly if not desirable?

xp-1000 (changes requested):

middleware/haproxy/detectors-haproxy.tf
cvauvarin: At first I thought this could be useful but after reading more documentation I  removed it.

xp-1000 (approved):
LGTM

PR #200, opened 2020-12-16
fix storage account capacity scale function definition

frederic-peraud:
UsedCapacity metric unit is Bytes. The scale function was wrong (`scale(1024*1024*1024)`)

### DIFF
--- a/modules/integration_azure-storage-account-capacity/detectors-azure-storage.tf
+++ b/modules/integration_azure-storage-account-capacity/detectors-azure-storage.tf
@@ -5,7 +5,7 @@ resource "signalfx_detector" "used_capacity" {
 
   program_text = <<-EOF
     base_filter = filter('resource_type', 'Microsoft.Storage/storageAccounts') and filter('primary_aggregation_type', 'true')
-    signal = data('UsedCapacity', filter=base_filter and ${module.filter-tags.filter_custom})${var.used_capacity_aggregation_function}${var.used_capacity_transformation_function}.scale(0.0000000001).publish('signal')
+    signal = data('UsedCapacity', filter=base_filter and ${module.filter-tags.filter_custom})${var.used_capacity_aggregation_function}${var.used_capacity_transformation_function}.scale(0.000000001).publish('signal')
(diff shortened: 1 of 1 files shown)


Shr3ps (changes requested):

BzSpi (approved):

xp-1000 (approved):

modules/integration_azure-storage-account-capacity/detectors-azure-storage.tf
```diff
@@ -3,7 +3,7 @@ resource "signalfx_detector" "used_capacity" {
 
   program_text = <<-EOF
     base_filter = filter('resource_type', 'Microsoft.Storage/storageAccounts') and filter('primary_aggregation_type', 'true')
-    signal = data('UsedCapacity', filter=base_filter and ${module.filter-tags.filter_custom})${var.used_capacity_aggregation_function}${var.used_capacity_transformation_function}.scale(1024*1024*1024).publish('signal')
+    signal = data('UsedCapacity', filter=base_filter and ${module.filter-tags.filter_custom})${var.used_capacity_aggregation_function}${var.used_capacity_transformation_function}.scale(0.000000001).publish('signal')
```
Shr3ps, line 6:
```suggestion
    signal = data('UsedCapacity', filter=base_filter and ${module.filter-tags.filter_custom})${var.used_capacity_aggregation_function}${var.used_capacity_transformation_function}.scale(0.0000000008).publish('signal')
```
to keep MiB unit

Shr3ps (changes requested):

modules/integration_azure-storage-account-capacity/detectors-azure-storage.tf
xp-1000: not a big fan but this is your choice. if you decide to apply this change so you have to update unit in rules descriptions

Shr3ps: Current unit in rule description is `GB` which represent octal base (8) and not decimal like done by the actual scale

Shr3ps (approved):

Shr3ps:
LGTM

PR #201, opened 2020-12-16
Switch detector to nagios plugin

arnnow:
Switch the detector to nagios.state instead of nagios_state.state.

This will work with the new monitor nagios activated on v5.7.0 of the agent : https://docs.signalfx.com/en/latest/integrations/agent/monitors/nagios.html

Also, enable runbook_url and tip for this detector.

### DIFF
--- a/modules/smart-agent_nagios-status-check/conf/readme.yaml
+++ b/modules/smart-agent_nagios-status-check/conf/readme.yaml
@@ -8,8 +8,8 @@ source_doc: |
 
   ### Monitors
 
-  You have to configure your existing Nagios script into the `nagios` monitor.
-  This will require agent >= [5.7.1](https://github.com/signalfx/signalfx-agent/releases/tag/v5.7.1)
+  This monitor is only available for agent `>= 5.7.1`. For prior versions, it is possible to use the 
+  `telegraf/exec` monitor with `nagios` parser but you will not get the script result in event.
   
   You can configure as many monitors as you have nagios checks to reuse but you have to define for each 
   it could be useful to adapt the `intervalSeconds` for each script.
@@ -17,21 +17,24 @@ source_doc: |
   ### Examples
 
   ```yaml
+  globalDimensions:
+    sfx_monitored: true
+
   - type: nagios
   intervalSeconds: 180
   command: /usr/lib/nagios/plugins/check_ntp_time -H pool.ntp.typhon.net -w 0.5 -c 1
   service: NTP_TIME
   extraDimensions:
-    sfx_monitored: true
+    sfx_monitored: false
 
   - type: nagios
   intervalSeconds: 180
   command: /usr/lib/nagios/plugins/check_http -I google.fr -f sticky -H google.fr -s http -u / -p 80
-  service: HTT_google
-  extraDimensions:
-    sfx_monitored: true
-
+  service: HTTP_google
   ```
+  
+  In this example, the first NTP check will not trigger any alert using default filtering convention but 
+  the metric and its value will be available.
 
 notes: |
   This module has been designed to alert with the same behavior as 
--- a/modules/smart-agent_nagios-status-check/conf/readme.yaml
+++ b/modules/smart-agent_nagios-status-check/conf/readme.yaml
@@ -1,47 +1,36 @@
 documentations:
   - name: Smart Agent monitor
-    url: 'https://docs.signalfx.com/en/latest/integrations/agent/monitors/telegraf-exec.html'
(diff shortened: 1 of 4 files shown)


modules/smart-agent_nagios-status-check/conf/readme.yaml
```diff
@@ -1,47 +1,28 @@
 documentations:
```
xp-1000, line 1: this doc is auto generated you have to edit its underlying yaml configuration in conf directory and run `make doc`

modules/smart-agent_nagios-status-check/variables.tf
```diff
@@ -50,3 +50,14 @@ variable "status_check_lasting_duration_seconds" {
   default     = "900"
 }
 
+variable "runbook_url" {
```
xp-1000, line 53: this is the object of another issue and so another PR which should be spread to the entire repo (not only one module) to keep modules homogeneous, same for `tip`

modules/smart-agent_nagios-status-check/conf/readme.yaml
```diff
@@ -51,5 +32,5 @@ notes: |
 
   While SignalFx does not provide `Unknown` severity this module uses the `Major` severity for unknown alerts.
 
-  The metric is named `nagios_state.state`, you need to add an `extraDimensions` to your monitor in order to be 
+  The metric is named `nagios.state`, you need to add an `extraDimensions` to your monitor in order to be
```
xp-1000, line 35: this is now false while monitor takes `service` as mandatory parameter

modules/smart-agent_nagios-status-check/conf/readme.yaml
```diff
@@ -1,47 +1,28 @@
...
-    telegrafParser:
-      dataFormat: nagios
-  - type: telegraf/exec
-    intervalSeconds: 900  
-    command: /usr/local/bin/scripts/check-megaraidsas
-    extraDimensions:
-      script: check_megaraidsas
-    telegrafParser:
-      dataFormat: nagios
+  - type: nagios
```
xp-1000, line 20: may be can you add more examples, like different use cases

modules/smart-agent_nagios-status-check/conf/readme.yaml
```diff
@@ -1,47 +1,28 @@
...
+  intervalSeconds: 180
+  command: /usr/lib/nagios/plugins/check_ntp_time -H pool.ntp.typhon.net -w 0.5 -c 1
+  service: NTP_TIME
+  extraDimensions:
+    sfx_monitored: false
```
xp-1000, line 25: it breaks the alerting so it must not be in the example (or at worst, as a counter example to explain how to collect metric without alert on this check)

xp-1000 (changes requested):

xp-1000:
we also to inform this module requires agent >= 5.7.1

modules/smart-agent_nagios-status-check/conf/readme.yaml
arnnow: I get an error when trying to generate the readme
```
terraform-signalfx-detectors:master ✓ ➭ make readmes modules/smart-agent_nagios-status-check/
./scripts/module/loop_wrapper.sh ./scripts/module/gen_doc.sh
sed: -e expression n°1, caractère 54: Fin d'intervalle invalide
make: *** [Makefile:34 : readmes] Erreur 1
```

modules/smart-agent_nagios-status-check/variables.tf
arnnow: Removed, i will try to create another PR for this

modules/smart-agent_nagios-status-check/conf/readme.yaml
arnnow: Removed from the doc

modules/smart-agent_nagios-status-check/conf/readme.yaml
arnnow: I added another exemple to the doc

modules/smart-agent_nagios-status-check/conf/readme.yaml
arnnow: yes, i switch `sfx_monitored` to true

arnnow:
> we also to inform this module requires agent >= 5.7.1

I did add this information to the readme. 
Tell me if that's ok with you

xp-1000 (approved):
Good, thanks

PR #202, opened 2020-12-18
#123 [Azure] Application Gateway - capacity detector

Shr3ps:


BzSpi (changes requested):
We should document how to find the `capacity_units_threshold_major` value since this concept is hardly understandable in my opinion.

Shr3ps:
> 
> 
> We should document how to find the `capacity_units_threshold_major` value since this concept is hardly understandable in my opinion.

done :)

BzSpi (approved):

xp-1000 (approved):

PR #208, opened 2020-12-21
Templating: Add unit on detectors messages #207

### LINKED ISSUE #207: Generator: add support for unit in detectors rules
In the detectors message, it would be fine to have the unit of the threshold

BzSpi:
Closes #207

### DIFF
--- a/modules/smart-agent_system-common/conf/03-space.yaml
+++ b/modules/smart-agent_system-common/conf/03-space.yaml
@@ -9,7 +9,9 @@ rules:
   critical:
     threshold: 90
     comparator: ">"
+    unit: "%"
   major:
     threshold: 80
     comparator: ">"
     dependency: critical
+    unit: "%"
--- a/modules/smart-agent_system-common/conf/03-space.yaml
+++ b/modules/smart-agent_system-common/conf/03-space.yaml
@@ -2,6 +2,7 @@ module: system
 name: "disk space utilization"
 id: disk_space
 transformation: ".max(over='5m')"
+value_unit: "%"
 signals:
   signal:
     metric: disk.utilization
@@ -9,9 +10,7 @@ rules:
   critical:
     threshold: 90
     comparator: ">"
-    unit: "%"
   major:
     threshold: 80
     comparator: ">"
     dependency: critical
-    unit: "%"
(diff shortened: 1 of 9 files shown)


xp-1000 (changes requested):
thanks, please could you document this in the https://github.com/claranet/terraform-signalfx-detectors/blob/master/scripts/templates/values.yaml ?

modules/smart-agent_system-common/conf/03-space.yaml
```diff
@@ -9,7 +9,9 @@ rules:
   critical:
     threshold: 90
     comparator: ">"
+    unit: "%"
```
xp-1000, line 12: why not define % for inodes too?

scripts/templates/variables.tf.j2
```diff
@@ -65,7 +65,7 @@ variable "{{ id }}_timeframe" {
   {% for key, rule in rules.items() -%}
     {%- set severity = key | lower -%}
 variable "{{ id }}_threshold_{{ severity }}" {
-  description = "{{ severity | capitalize }} threshold for {{ id }} detector"
+  description = "{{ severity | capitalize }} threshold for {{ id }} detector{% if rule.unit | default(false) %} (unit: {{ rule.unit }}){% endif %}"
```
xp-1000, line 68:
```suggestion
  description = "{{ severity | capitalize }} threshold for {{ id }} detector{% if rule.unit is defined %} in {{ rule.unit }}{% endif %}"
```

scripts/templates/values.yaml
```diff
@@ -154,3 +154,8 @@ rules:
     ## This will add another condition to negate the dependecy like:
     ## when(signal > {{severity}}) and when(signal <= {{severity.dependency}})
     #dependency:
+
+    ## @param unit - string - optional
+    ## Threshold unit for the current detector.
+    ## This will update the detector message and associated variables description
+    unit:
```
xp-1000, line 161: I am not sure it is the good place to put it.
Define it at detector level will be easier to configure and more suited to our model where one detector is atomic and only contains 1 "check"

xp-1000 (changes requested):

xp-1000 (approved):

PR #209, opened 2020-12-21
Azure: Add global Storage Accounts detectors

### LINKED ISSUE #206: Cloud Azure - Storage detectors
Rework Azure Storage Account detectors to replace those in the stable branch.

Proposal: 6 new modules
* general storage account with Azure limits
* blob storage
* queue storage
* table storage
* file standard
* file premium (due to different behavior)

Claranet reference: AZ-394

BzSpi:
Adding global storage account detector. 
Checks:
* overall capacity limit
* resources count limit
* ingress limit
* egress limit
* transactions rate limit

fixes #206

### DIFF
--- a/modules/integration_azure-storage-account/conf/01-count.yaml
+++ b/modules/integration_azure-storage-account/conf/01-count.yaml
@@ -1,0 +1,19 @@
+module: "Azure Storage Account"
+name: "count"
+filtering: "filter('resource_type', 'Microsoft.Storage/storageAccounts') and filter('primary_aggregation_type', 'true')"
+aggregation: ".count(by=['subscription_id'])"
+transformation: ".min(over='1d')"
+signals:
+  capacity:
+    metric: "UsedCapacity"
+  signal:
+    formula:
+      capacity.fill(None, duration='1d')
+rules:
+  critical:
+    threshold: 245
+    comparator: ">"
+  major:
+    threshold: 240
+    comparator: ">"
+    dependency: critical
--- a/modules/integration_azure-storage-account/variables-gen.tf
+++ b/modules/integration_azure-storage-account/variables-gen.tf
@@ -1,0 +1,310 @@
+# count detector
+
+variable "count_notifications" {
+  description = "Notification recipients list per severity overridden for count detector"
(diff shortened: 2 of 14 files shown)


Shr3ps (approved):

modules/integration_azure-storage-account/conf/01-count.yaml
```diff
@@ -0,0 +1,16 @@
+module: "Azure Storage Account"
+name: "count"
+filtering: "filter('resource_type', 'Microsoft.Storage/storageAccounts') and filter('primary_aggregation_type', 'true')"
+aggregation: ".fill(None, duration='1d').count(by=['subscription_id'])"
```
xp-1000, line 4: why not simply `fill()` ?

modules/integration_azure-storage-account/variables-gen.tf
```diff
@@ -0,0 +1,310 @@
...
+variable "capacity_notifications" {
+  description = "Notification recipients list per severity overridden for capacity detector"
+  type        = map(list(string))
+  default     = {}
+}
+
+variable "capacity_aggregation_function" {
+  description = "Aggregation function and group by for capacity detector (i.e. \".mean(by=['host'])\")"
+  type        = string
+  default     = ".fill(None, duration='1d').sum(by=['azure_resource_name', 'azure_resource_group_name', 'azure_region'])"
```
xp-1000, line 74: I understand why you do this. while it is not possible to use fill in current generator template you can trick it here.

I think a better approach would be to use a formula base on this simple metric signal and use fill inside this formula (and not the underlying metric signal).

xp-1000 (changes requested):

modules/integration_azure-storage-account/conf/01-count.yaml
BzSpi: Because I want it to stop after one day :)

xp-1000: ok, it is not common so I would to be sure :p

xp-1000 (approved):

PR #212, opened 2020-12-22
#211 Generator: add rollup feature on signals

### LINKED ISSUE #211: Generator: add support for rollup in signals
Like extrapolation, we should support rollup in signals.

BzSpi:
Add the `rollup` function for signals development with generator.

Fixes #211

### DIFF
--- a/scripts/templates/detector.tf.j2
+++ b/scripts/templates/detector.tf.j2
@@ -32,9 +32,9 @@ resource "signalfx_detector" "{{ id }}" {
 
     {%- for key, signal in signals.items() -%}
       {%- if 'metric' in signal %}
-    {{ key }} = data('{{ signal.metric }}', filter={%- if exclude_not_running_vm is defined and exclude_not_running_vm -%}${local.not_running_vm_filters} and {% endif -%}{%- if filtering is defined and filtering -%}{{ filtering_variable }} and {% endif -%}${module.filter-tags.filter_custom}{%- if signal.extrapolation is defined and signal.extrapolation -%}, extrapolation='{{ signal.extrapolation }}'{% endif -%})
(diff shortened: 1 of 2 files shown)


scripts/templates/detector.tf.j2
```diff
@@ -32,9 +32,9 @@ resource "signalfx_detector" "{{ id }}" {
 
     {%- for key, signal in signals.items() -%}
       {%- if 'metric' in signal %}
-    {{ key }} = data('{{ signal.metric }}', filter={%- if exclude_not_running_vm is defined and exclude_not_running_vm -%}${local.not_running_vm_filters} and {% endif -%}{%- if filtering is defined and filtering -%}{{ filtering_variable }} and {% endif -%}${module.filter-tags.filter_custom}{%- if signal.extrapolation is defined and signal.extrapolation -%}, extrapolation='{{ signal.extrapolation }}'{% endif -%})
+    {{ key }} = data('{{ signal.metric }}', filter={%- if exclude_not_running_vm is defined and exclude_not_running_vm -%}${local.not_running_vm_filters} and {% endif -%}{%- if filtering is defined and filtering -%}{{ filtering_variable }} and {% endif -%}${module.filter-tags.filter_custom}{%- if signal.extrapolation is defined and signal.extrapolation -%}, extrapolation='{{ signal.extrapolation }}'{% endif -%}{%- if signal.rollup is defined -%}, rtype='{{ signal.rollup }}'{% endif -%})
```
xp-1000, line 35: `rtype=` -> `rollup=` ?

we already use `rollup` in this repo and we use `filter` (not `fobject`) so it will be more consistent and understandable.

xp-1000 (changes requested):
one day a friend of mine explained to me that a good naming is more important than a good doc :p

scripts/templates/detector.tf.j2
xp-1000: also it will be better to add the rollup *before* the extrapolation as the signalfx builder does (it will avoid undesirable diff after edit from webui)

xp-1000 (approved):

Issue #105, opened 2020-09-24
Common submodule detector


xp-1000:
this one will be hard and will be a huge change on the repo but also the way we create detectors.

currently we defined detector resource https://www.terraform.io/docs/providers/signalfx/r/detector.html directly inside each module which lead to much duplicated code and a contribution process little tedious.

We already have common module with `filter-tags`, the goal is to create another one dedicated to detector creation templating.

Each detector created in modules will be created calling the template submodule which could: 

- enforce good practices
- ease the contribution adding detectors
- add ability to add some complex HCL stuff without to duplicate this to every modules (add locals ...)
- greatly facilitate changes in repo avoiding to update code in lot of modules, only one location to updated will affect every detectors.

PR #219, opened 2020-12-23
elasticache: lower free memory alerts severity

NSenaud:
Detector checks for fast free memory decrease rather than actual low memory on the node, so false-positive are more than likely and a Critical severity seems a bit high in this regards.

This PR converts Critical rule to Major, and Major's to Minor.

### DIFF
--- a/modules/integration_aws-elasticache-common/detectors-elasticache.tf
+++ b/modules/integration_aws-elasticache-common/detectors-elasticache.tf
@@ -147,21 +147,9 @@ resource "signalfx_detector" "free_memory" {
 
   program_text = <<-EOF
     signal = data('FreeableMemory', filter=filter('namespace', 'AWS/ElastiCache') and filter('stat', 'lower') and filter('CacheNodeId', '*') and ${module.filter-tags.filter_custom}).rateofchange()${var.free_memory_aggregation_function}${var.free_memory_transformation_function}.publish('signal')
-    detect(when(signal < ${var.free_memory_threshold_critical})).publish('CRIT')
-    detect(when(signal < ${var.free_memory_threshold_major}) and when(signal >= ${var.free_memory_threshold_critical})).publish('MAJOR')
+    detect(when(signal < ${var.free_memory_threshold_major})).publish('MAJOR')
+    detect(when(signal < ${var.free_memory_threshold_minor}) and when(signal >= ${var.free_memory_threshold_major})).publish('MINOR')
 EOF
-
-  rule {
-    description           = "is too low < ${var.free_memory_threshold_critical}"
-    severity              = "Critical"
-    detect_label          = "CRIT"
-    disabled              = coalesce(var.free_memory_disabled_critical, var.free_memory_disabled, var.detectors_disabled)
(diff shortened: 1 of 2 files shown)


xp-1000 (changes requested):
thanks @NSenaud 

I was about to merge this but may be is there a better way to do this.

we could remove this detector from the `common` module and add it to both modules:
- `memcached` using `BytesUsedForCacheItems` cloudwatch metric
- `redis` using `BytesUsedForCache` cloudwatch metric 

alongs with `FreeableMemory` metric we should be able to calculate a % memory usage as usually. Even better for redis only there is the `DatabaseMemoryUsagePercentage` ready to use


Reading other detectors, I think there are lot of easy improvements to do here. 

The main problem comes from the many possible uses of this services: cache, storage, queue, redis/memcached and the "resilience"/kind of workload behind.
If we really want to be generic so we should only have CPU detector (and heartbeat). All other detectors are as much relevant and useful in some cases as false positive and useless in others.

The best way would be to have an "abstraction" like multiple modules which call the common one with different parameters to cover different popular use cases (cache, storage..). For example, disable evictions for cache (but keeping evictions growing) and the reverse for storage use case. But we do not have implemented this logic yet, even if it seems feasible it is significant work and should be in the scope of the v2 https://github.com/claranet/terraform-signalfx-detectors/milestone/2 especially https://github.com/claranet/terraform-signalfx-detectors/issues/105.

So for now, what I propose is to choose the most popular use case and adapt the default config to fit it. If the need to cover another use case comes before to release v2 so we still could simply duplicate this module waiting for (but obviously I prefer to avoid while possible).

That said we could add some notes to the readme to explain what is possible for the user to configure to switch to other use case like "disable this detector when use redis as storage but enable this one ...".

What do you think @NSenaud ?

Here I list some changes to do if we decide to make the current modules (common,redis,memcached) fit the "cache" use case (i.e. for web app).

- I would disable evictions detector because the only worrying thing about evictions here would be progressive grow. It could remain enable but in minor/warning because it is rarely a problem itself, it is often useful to know. In any case even if not generic is is useful for those who know well their application and can define an abnormal threshold so we should not delete it.
- swap I could change the threshold "50MB" for major and at least x2 for critical to follow the aws recommendation (but mitigate for critical because they prefer you pay bigger node). I would also add a `tip` to explain when swap happen you should first check the configured reserved memory in parameter group https://docs.aws.amazon.com/AmazonElastiCache/latest/red-ug/redis-memory-management.html and upscale after.
- the max connection must be disable. it could be useful but the user has to configure it properly. may be a `tip` to explain the limit depends on the instance type (or the parameter group if explicitly configured).
- I think no connection is good especially for cache the app should be always connected (and we can expect at least one client) but I would lower to major instead of critical
- evictions growing is good for cache, evictions is normal but progressive growing is fatally a problem (or as soon as the hit rate is bad) but it is not an availability problem so -> major max
- memory has been detailed at the start

of course we can also add more detectors like replication or more complex signaflow like historical anomaly etc. this comment focus on remain on the same scope but just give it sens (with an opinionated "cache" choice of elasticache use)

xp-1000 (approved):
we will merge this as it is and keep trace of suggestions in issue https://github.com/claranet/terraform-signalfx-detectors/issues/225

PR #220, opened 2020-12-23
#117 add authorized_writer_teams to generator and common vars

### LINKED ISSUE #117: Permissions and team management
we need to be able to configure permissions on detectors

xp-1000:
close #117


Shr3ps (commented):
Does this attribute is available with any version of the provider? (do we need to update versions.tf or not?)

xp-1000:
> Does this attribute is available with any version of the provider? (do we need to update versions.tf or not?)

no need to upgrade the provider

Shr3ps (approved):

PR #226, opened 2021-01-05
Hearbeat detector for Azure Container Instance

markyam:
Init Azure Container Instance module with hearbeat detector.


modules/integration_azure-container-instance/conf/readme.yaml
```diff
@@ -0,0 +1,3 @@
+documentations:
```
xp-1000, line 1: fillme fillme !

modules/integration_azure-container-instance/conf/readme.yaml
```diff
@@ -0,0 +1,3 @@
...
+
+source_doc:
```
xp-1000, line 3: may be add `notes` to warn about the risk to get false alert for "temp"/"short lived" containers ?

xp-1000 (changes requested):

modules/integration_azure-container-instance/conf/readme.yaml
markyam: done

modules/integration_azure-container-instance/common-modules.tf
```diff
@@ -0,0 +1 @@
+../../common/module/modules.tf
```
Shr3ps, line 1: this file must be deleted, and you should use the Azure specific one (with azure tags)

modules/integration_azure-container-instance/conf/00-hearbeat.yaml
```diff
@@ -0,0 +1,11 @@
+module: Azure Container Instance
+name: heartbeat
+filtering: "filter('resource_type', 'Microsoft.ContainerInstance/containerGroups') and filter('primary_aggregation_type', 'true')"
+transformation: false
+aggregation: true
```
Shr3ps, line 5: This should be a mean by name, rg, region

Shr3ps (changes requested):

modules/integration_azure-container-instance/conf/00-hearbeat.yaml
xp-1000: I don't think this what he wants.

In my opinion this seems far safer to group by something "above" a container because in my mind a container is a short lived / dynamic resource where disappearing should not be a cause of legitimate alert.

but if I right understood he wants to have an alert as soon as container disappears. In this case, using group by breaks this BUT it will require a note about this (see https://github.com/claranet/terraform-signalfx-detectors/pull/226#discussion_r552024954)

modules/integration_azure-container-instance/common-modules.tf
markyam: File replaced by the Azure one

modules/integration_azure-container-instance/conf/00-hearbeat.yaml
markyam: Aggregation function added

xp-1000 (approved):

Shr3ps (approved):

modules/integration_azure-container-instance/conf/00-hearbeat.yaml
```diff
@@ -0,0 +1,11 @@
+module: Azure Container Instance
+name: heartbeat
+filtering: "filter('resource_type', 'Microsoft.ContainerInstance/containerGroups') and filter('primary_aggregation_type', 'true')"
+transformation: false
+aggregation: ".mean(by=['azure_resource_name', 'azure_resource_group_name', 'azure_region'])"
```
BzSpi, line 5: Would aggregation by `containerName` be a good idea ? I don't really know about ACI but dimension seems available.

modules/integration_azure-container-instance/conf/readme.yaml
```diff
@@ -0,0 +1,11 @@
+documentations:
+  - name: Azure Monitor metrics
+    url: 'https://docs.microsoft.com/en-us/azure/azure-monitor/platform/metrics-supported'
```
BzSpi, line 3:
```suggestion
    url: 'https://docs.microsoft.com/en-us/azure/azure-monitor/platform/metrics-supported#microsoftcontainerinstancecontainergroups'
```

modules/integration_azure-container-instance/conf/00-hearbeat.yaml
xp-1000: I think it will not have any impact on alert body but it will add this to dimensions list in its subject.

so it seems to be a good idea but it could be undesired if `azure_resource_name == containerName`.

This is up to the author

modules/integration_azure-container-instance/conf/readme.yaml
xp-1000: why not but if you apply this suggestion you should remove this link from the `source_doc`, not useful to duplicate the information.

Keep it as it is makes sens for me : 
- related doc list (here) uses the azure monitor source, global documentation
- additional info about source includes a more specific link to the right resource (source_doc)

not obvious in the config file but fully logic on the rendered reamde.md

modules/integration_azure-container-instance/conf/00-hearbeat.yaml
markyam: ACI is a container group so resources are shared between them. The heartbeat detector is about a metric from all containers of the ACI so I don't think aggregation by `containername` would be pertinent.

BzSpi (approved):

PR #227, opened 2021-01-05
Add nodetool check detector

Gati0:
This PR proposes to adds 2 detectors. These detectors verify that Cassandra node is up and running in the cluster.
This is the equivalent of a "nodetool status".
The metrics are reported via the signalfx agent and a groovy script (visible here: https://docs.signalfx.com/en/latest/integrations/agent/monitors/jmx.html)

For information, the metrics can return the following values:
- cassandra.status: 1 == Live, 0 == Dead, -1 == Unknown
- cassandra.state: 1 == Normal, 2 == Leave, 3 == Join

If the code "1" is not received for each metrics, then an alert will be triggered.

### DIFF
--- a/modules/smart-agent_cassandra-nodetool/conf/00-status.yaml
+++ b/modules/smart-agent_cassandra-nodetool/conf/00-status.yaml
@@ -1,0 +1,16 @@
+module: cassandra nodetool
+name: "Node status"
+transformation: ".min(over='10m')"
+aggregation: true
+signals:
+  signal:
+    metric: "cassandra.status"
+rules:
+  critical:
+    threshold: 0
+    comparator: "=="
+    description: "is dead"
+  minor:
+    threshold: 0
+    comparator: "<"
+    description: "is unknown"
--- a/modules/smart-agent_cassandra-nodetool/conf/01-state.yaml
+++ b/modules/smart-agent_cassandra-nodetool/conf/01-state.yaml
@@ -1,0 +1,13 @@
+module: cassandra nodetool
+name: "Node state"
+transformation: ".min(over='1h')"
+aggregation: true
+signals:
+  signal:
+    metric: "cassandra.state"
+rules:
+  critical:
+    threshold: 1
+    comparator: ">"
+    description: "is not normal"
+tip: "The node state may be in leaving/joining for too long. Check the nodetool status result"
--- a/modules/smart-agent_cassandra-nodetool/conf/readme.yaml
+++ b/modules/smart-agent_cassandra-nodetool/conf/readme.yaml
@@ -4,10 +4,15 @@ documentations:
 
 source_doc: |
 
-  This detectors requires a groovy script to get the metrics `cassandra.status` and `cassandra.state`. The script can be found in the `scripts` folder of this module.
-  `cassandra.status` and `cassandra.state` are metrics that track the state of the node in the cluster. The nodetool status command returns the equivalent information.
+  This module leverages the powerful `JMX` monitor to retrieve equivalent information 
+  from the output of [Cassandra 
+  Nodetool](https://cassandra.apache.org/doc/latest/tools/nodetool/nodetool.html) like 
+  the `status` and the `state` of a node and push them as metrics.
 
-  For example:
+  The advantage of this method is we do not need to make the agent able to run `nodetool` 
+  command on Cassandra nodes (i.e. by configuring `sudo`) and rely on an unsafe output 
+  parsing:
+
   ```
   nodetool status mykeyspace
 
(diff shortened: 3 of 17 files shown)


xp-1000 (changes requested):
given that this is a different source of data I would create a new separate module for nodetool like `make module smart-agent_cassandra-nodetool`

modules/smart-agent_cassandra-nodetool/conf/00-status.yaml
```diff
@@ -0,0 +1,14 @@
+## Example
+module: smart-agent_cassandra-nodetool
```
xp-1000, line 2:
```suggestion
module: "Cassandra Nodetool"
```

modules/smart-agent_cassandra-nodetool/conf/00-status.yaml
```diff
@@ -0,0 +1,18 @@
+## Example
+module: cassandra nodetool
+name: "Node is not in live status"
+transformation: ".min(over='10m')"
+
+aggregation: false
```
xp-1000, line 6: I would set it to true to give the user ability to cusotmize this from variable if needed:

```suggestion
aggregation: true
```

modules/smart-agent_cassandra-nodetool/conf/01-state.yaml
```diff
@@ -0,0 +1,14 @@
+## Example
+module: cassandra nodetool
+name: "Node is not in normal status"
+
+transformation: ".min(over='10m')"
+aggregation: false
```
xp-1000, line 6: I would set it to true to give the user ability to cusotmize this from variable if needed:

```suggestion
aggregation: true
```

modules/smart-agent_cassandra-nodetool/conf/01-state.yaml
```diff
@@ -0,0 +1,14 @@
+## Example
+module: cassandra nodetool
+name: "Node is not in normal status"
+
+transformation: ".min(over='10m')"
```
xp-1000, line 5: I am not sure of the time it can take for a node to join/leave "normally" a cluster.

may be it is safer to increase a bit this timeframe?

modules/smart-agent_cassandra-nodetool/conf/readme.yaml
```diff
@@ -0,0 +1,3 @@
+documentations:
```
xp-1000, line 1:
```suggestion
documentations:
  - name: Smart Agent monitor
    url: 'https://docs.signalfx.com/en/latest/integrations/agent/monitors/jmx.html'
```

modules/smart-agent_cassandra-nodetool/conf/readme.yaml
```diff
@@ -0,0 +1,3 @@
...
+
+source_doc:
```
xp-1000, line 3: you should inform it requires to use the groovy script, probably add this script to git versionning in this module directory and ideally provide a sample agent config

xp-1000 (changes requested):

modules/smart-agent_cassandra-nodetool/conf/00-status.yaml
```diff
@@ -0,0 +1,18 @@
+## Example
```
xp-1000, line 1: You can drop this line while it is not an example.

also there is no explicit rule about that but in general we use the `00` for the heartbeat so if it is ok for you please start count at 1 ?

modules/smart-agent_cassandra-nodetool/conf/00-status.yaml
```diff
@@ -2,17 +2,12 @@
 module: cassandra nodetool
 name: "Node is not in live status"
```
xp-1000, line 3:
```suggestion
name: "Node status"
```

modules/smart-agent_cassandra-nodetool/conf/00-status.yaml
```diff
@@ -0,0 +1,13 @@
+## Example
+module: cassandra nodetool
+name: "Node is not in live status"
+transformation: ".min(over='10m')"
+aggregation: false
+signals:
+  signal:
+    metric: "cassandra.status"
+rules:
+  critical:
```
xp-1000, line 10: I would keep 2 rules different to have 2 different severities for different "abnormal" statuses.

You was closer with before last change:
- you just had to customize the description like `is dead` && `is unknown` keeping `<` and rules dependency as it was.

but you can keep the new more "strict" logic:
- removing dependency between them just, a different severity like `critical` for dead and `minor` for unknown
- then you can use `== 0` && `== -1` for dead and unknown
- as for the first solution : update the description now it is possible

modules/smart-agent_cassandra-nodetool/conf/01-state.yaml
xp-1000: Also, a single node down is not as much problematic than the cluster down.

here there is no cluster aggregation by default so I would set a long value here because:
- a member in abnormal state during long time is ALWAYS a problem (>= 1hour I think)
- a member in abnormal state during few time is ALWAYS NOT a problem

xp-1000 (changes requested):

xp-1000 (approved):

PR #232, opened 2021-01-06
 Azure: Add Storage Accounts premium file detectors

BzSpi:
Relates to #206


modules/integration_azure-storage-file-premium/common-versions.tf
```diff
@@ -0,0 +1,9 @@
+terraform {
```
Shr3ps, line 1: why not symbolic link for those kind of files?

Shr3ps (changes requested):

modules/integration_azure-storage-file-premium/common-versions.tf
BzSpi: Wrong copy/paste :)

BzSpi:
Set the non working detectors (those without filtering) to disabled by default.

PR #234, opened 2021-01-08
Add severity doc and gen script

cvauvarin:
Generate severity doc per detector.


Makefile
```diff
@@ -1,4 +1,4 @@
-SUPPORTED_COMMANDS := init-module update-module update-module-doc update-module-tf update-module-detectors update-module-outputs check-module
+SUPPORTED_COMMANDS := init-module update-module update-module-doc update-module-tf update-module-detectors update-module-outputs update-severity-doc check-module
```
xp-1000, line 1: you should not add the target here while we do not want to support any argument for this target

xp-1000 (changes requested):
just a little comment but it seems fine and could be merged as soon as the severity doc will be added.

xp-1000 (approved):

PR #239, opened 2021-01-14
add teams common variable

frederic-peraud:
This PR adds `Teams` as a common parameter. It allows to link detector(s) to Sfx Teams.

https://registry.terraform.io/providers/splunk-terraform/signalfx/latest/docs/resources/detector#teams

### DIFF
--- a/common/module/variables.tf
+++ b/common/module/variables.tf
@@ -52,3 +52,15 @@ variable "authorized_writer_teams" {
   default     = null
 }
 
+
+variable "authorized_writer_teams" {
+  description = "List of teams IDs authorized (with admins) to edit the detector. If defined, it requires an user token to work"
+  type        = list(string)
+  default     = null
+}
+
+variable "teams" {
+  description = "List of teams IDs to associate the detector to"
+  type        = list(string)
+  default     = null
+}
--- a/common/module/variables.tf
+++ b/common/module/variables.tf
@@ -55,5 +55,5 @@ variable "authorized_writer_teams" {
 variable "teams" {
   description = "List of teams IDs to associate the detector to"
   type        = list(string)
-  default     = null
+  default     = []
 }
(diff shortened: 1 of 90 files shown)


xp-1000:
thanks @frederic-peraud

I pushed atomic commits to detail all steps to make this kind of change:
- add the variable declaration, this is what you already did. it will be obviously more painful if should declared at detector level.
- manually edit all "handmade" detectors tf files (no `-gen` in names) to use the variable. I use a crappy bash command for that: `find -type f -name 'detectors-*.tf' ! -name 'detectors-gen.tf' -exec sed -E -i -r 's/^( *name *= format.*)$/\1\n\n  toto = var.toto/' {} \;` it will add toto definition to all detectors (not generated). [`aae48b4` (#239)](https://github.com/claranet/terraform-signalfx-detectors/pull/239/commits/aae48b49953bb5b65066c04278631012861c707d)
- then you will have to update the jinj2 template generator to add this definition, it is generally easy except that we have to be careful on keep a rendered output terraform fmt-ed in all scenarios (depending on yaml conf). [`909b442` (#239)](https://github.com/claranet/terraform-signalfx-detectors/pull/239/commits/909b442b67ea3faba45578bf08a57822f9e5fa57)
- Finally run `make update-module` to let the generator do the job for the rest of files: [`83af1ad` (#239)](https://github.com/claranet/terraform-signalfx-detectors/pull/239/commits/83af1adefb38ef9375ae47d50a9931e96a021ef6)

the value assigned to `teams` attribute and the last commit propose to : 
- take the `authorized_writer_teams` value if it is defined but not `teams`
- take the `teams` value only if defined (not matter if `authorized_writer_teams` is defined or not).

It is only a proposition, if you disagree simply reset the last commit and update the definition of `teams` in all detectors (and the generator) to simply use the `teams` variable (remove try, coalescelist and `authorized_writer_teams`)

frederic-peraud:
Hi @xp-1000 

Your suggestions are ok for me and my team.

xp-1000 (approved):

Shr3ps (approved):

PR #241, opened 2021-01-25
kubernetes-common: fix OMMKilled detector

NSenaud:
Add .max(over='1m') to avoid triggering alert repeatedly

### DIFF
--- a/modules/smart-agent_kubernetes-common/README.md
+++ b/modules/smart-agent_kubernetes-common/README.md
@@ -14,6 +14,7 @@
 - [Notes](#notes)
   - [About `node_ready` detector](#about-node_ready-detector)
   - [About `job_failed` detector](#about-job_failed-detector)
+  - [About `oom_killed` detector](#about-oom_killed-detector)
 - [Related documentation](#related-documentation)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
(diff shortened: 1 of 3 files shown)


modules/smart-agent_kubernetes-common/detectors-kubernetes-common.tf
```diff
@@ -110,7 +110,7 @@ resource "signalfx_detector" "oom_killed" {
   authorized_writer_teams = var.authorized_writer_teams
 
   program_text = <<-EOF
-    signal = data('kubernetes.container_ready', filter=filter('container_status', 'terminated') and filter('container_status_reason', 'OOMKilled') and ${module.filter-tags.filter_custom})${var.oom_killed_aggregation_function}${var.oom_killed_transformation_function}.count().publish('signal')
+    signal = data('kubernetes.container_ready', filter=filter('container_status', 'terminated') and filter('container_status_reason', 'OOMKilled') and ${module.filter-tags.filter_custom})${var.oom_killed_aggregation_function}${var.oom_killed_transformation_function}.count().max(over='1m').publish('signal')
```
xp-1000, line 113: I prefer you use the available `oom_killed_transformation_function` variable and set a default value instead of hardcode this here

xp-1000 (changes requested):

xp-1000 (changes requested):
seems good thanks

feel free to add a note about this in the readme it could be valuable to understand why this default value

xp-1000 (approved):

NSenaud:
Great! I added a note in README, rebased the branch and squashed commits.

PR #243, opened 2021-01-25
adding some detectors for azure storage

nicolasvion:
adding some detectors for azure storage

### DIFF
--- a/modules/integration_azure-storage-account/conf/03-ingress.yaml
+++ b/modules/integration_azure-storage-account/conf/03-ingress.yaml
@@ -7,7 +7,7 @@ value_unit: "Gbps"
 signals:
   ingress:
     metric: "Ingress"
-    rollup: rate
+    rollup: sum
   signal:
     formula:
       ingress.scale(0.000000008) # Scale to Gb unit
--- a/modules/integration_azure-storage-account/conf/06-requests-rate-status.yaml
+++ b/modules/integration_azure-storage-account/conf/06-requests-rate-status.yaml
@@ -1,16 +1,15 @@
 module: "Azure Storage Account"
 name: "requests rate status"
-filtering: "filter('resource_type', 'Microsoft.Storage/storageAccounts') and filter('primary_aggregation_type', 'true')"
 aggregation: ".sum(by=['azure_resource_name', 'azure_resource_group_name', 'azure_region'])"
 transformation: ".min(over='5m')"
 signals:
   rate_success:
     metric: "Transactions"
     filtering: "filter('resource_type', 'Microsoft.Storage/storageAccounts') and filter('responsetype', 'Success') and filter('primary_aggregation_type', 'true')"
-    rollup: rate
+    rollup: sum
   rate_failed:
     metric: "Transactions"
-    rollup: failed
+    rollup: sum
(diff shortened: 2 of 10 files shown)


modules/integration_azure-storage-account/conf/03-ingress.yaml
```diff
@@ -7,7 +7,7 @@ value_unit: "Gbps"
 signals:
   ingress:
     metric: "Ingress"
-    rollup: rate
+    rollup: sum
```
BzSpi, line 10: Why ? To have the value per second, we need a rate aggregation.

modules/integration_azure-storage-account/conf/06-requests-rate-status.yaml
```diff
@@ -0,0 +1,25 @@
+module: "Azure Storage Account"
+name: "requests rate status"
+filtering: "filter('resource_type', 'Microsoft.Storage/storageAccounts') and filter('primary_aggregation_type', 'true')"
+aggregation: ".sum(by=['azure_resource_name', 'azure_resource_group_name', 'azure_region'])"
+transformation: ".min(over='1m')"
+signals:
+  rate_success:
+    metric: "Transactions"
+    filter: "filter('responsetype', 'Success')"
```
BzSpi, line 9: Why filter by status then sum ? We can avoid filtering so that we have the sum.

modules/integration_azure-storage-account/conf/06-requests-rate-status.yaml
```diff
@@ -0,0 +1,25 @@
+module: "Azure Storage Account"
+name: "requests rate status"
+filtering: "filter('resource_type', 'Microsoft.Storage/storageAccounts') and filter('primary_aggregation_type', 'true')"
+aggregation: ".sum(by=['azure_resource_name', 'azure_resource_group_name', 'azure_region'])"
+transformation: ".min(over='1m')"
```
BzSpi, line 5: One minute is way too short. We will have a lot of instability.

modules/integration_azure-storage-account/conf/06-requests-rate-status.yaml
```diff
@@ -0,0 +1,25 @@
+module: "Azure Storage Account"
```
BzSpi, line 1: having a global requests status detector does not make sense in my opinion (we have removed it from our previous Datadog implementation). We should monitor the failed request rate per usage (blob, table ...).

modules/integration_azure-storage-account/conf/07-latency-e2e.yaml
```diff
@@ -0,0 +1,17 @@
+module: "Azure Storage Account"
```
BzSpi, line 1: Having a global latency status detector does not make sense in my opinion. We may have big variations depending if we use tables or files.

BzSpi (commented):
I do not agree with new suggested detectors and current detectors changes. 
Current detectors work as expected and new detectors should be implemented per usage (blob, table...).
I'm open to discuss about it.

nicolasvion:
Internal conversation, closing this MR. Creating new one for the specific detectors on storage object (blob, etc...)

PR #244, opened 2021-02-03
adding vars to allow to disable detector in generator mode

nicolasvion:
- adding vars to allow to disable detector in generator mode
- removing trailing spaces

### DIFF
--- a/scripts/templates/values.yaml
+++ b/scripts/templates/values.yaml
@@ -11,27 +11,27 @@
 ## Use "-" as separator, no spaces or underscores.
 ## Aims to be passed from env var by script or cli.
 #
-module: 
+module:
 
 ## @param id - string - optional
 ## Detector id (shorter canonical name).
 ## Used as prefix for terraform variables, resources and outputs names.
 #
 ## If not defined it will result in lower {{name}} with spaces replaced by `_`.
-## This is why this is highly recommended to explicitly set {{id}} if 
+## This is why this is highly recommended to explicitly set {{id}} if
 ## {{name}} is too long or use special chars like `()`.
 #
 ## Aims to be passed from env var by script or cli.
 #
-id: 
+id:
 
 ## @param name - string - mandatory
 ## Detector entire readable name (what it checks).
 ## Used after {{module}} for alert subject and as prefix for variables.
 #
-## The name will determine the detector type to create. If equals to 
+## The name will determine the detector type to create. If equals to
 ## "heartbeat", it will use "not_reporting.detector()" function and add
-## "max_delay=900". Else, it writes threshold detector type using 
+## "max_delay=900". Else, it writes threshold detector type using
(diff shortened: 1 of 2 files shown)


scripts/templates/variables.tf.j2
```diff
@@ -53,7 +53,7 @@ variable "{{ id }}_runbook_url" {
 variable "{{ id }}_disabled" {
   description = "Disable all alerting rules for {{ id }} detector"
   type        = bool
-  default     = null
+  default     = {% if disabled %}{{ disabled }}{% else %}null{% endif %}
```
Shr3ps:
```suggestion
  default     = {% if disabled is not None %}{{ disabled }}{% else %}null{% endif %}
```

?

if we set disabled = false, this test will put "null", no?

Shr3ps (changes requested):

scripts/templates/values.yaml
```diff
@@ -89,23 +89,28 @@ value_unit:
...
 ## Define a default tip for all rules of the detector, see:
 ## https://docs.signalfx.com/en/latest/detect-alert/set-up-detectors.html#tip
-## If not defined, an empty string will be used as variable to allow the 
+## If not defined, an empty string will be used as variable to allow the
 ## use to set its custom one.
-tip: 
+tip:
+
+## @param disabled - bool - optional
+## Disabled parameters for the current detector
```
xp-1000, line 104: I think you enrich this description to explain the full behavior like:

```
Allows to disable all alerting rules defined in this detector 
it works only if set to `true`.
```

scripts/templates/variables.tf.j2
```diff
@@ -53,7 +53,7 @@ variable "{{ id }}_runbook_url" {
 variable "{{ id }}_disabled" {
   description = "Disable all alerting rules for {{ id }} detector"
   type        = bool
-  default     = null
+  default     = {% if disabled is defined %}{{ disabled }}{% else %}null{% endif %}
```
xp-1000, line 56: I don't think it is explicit/safe enough.
the values.yaml file is provided to be copyable and works without modification.

as you define this in values.yaml so it will disable all alerting by default and this is not the desirable behavior I think.

```suggestion
  default     = {% if disabled is boolean and disabled is sameas true %}true{% else %}null{% endif %}
```

xp-1000 (changes requested):

scripts/templates/variables.tf.j2
xp-1000: it seems we do not need the first test so this is enough: 

```suggestion
  default     = {% if disabled is sameas true %}true{% else %}null{% endif %}
```

Shr3ps (approved):

xp-1000 (approved):
Good, thanks!

PR #245, opened 2021-02-11
init blob detectors - disabled by default, need explicit activation

nicolasvion:
- adding blob storage detectors
- disabled by default, need explicit activation because of some stability pb
- customers' request

### DIFF
--- a/modules/integration_azure-storage-account-blob/conf/01-requests-rate-status.yaml
+++ b/modules/integration_azure-storage-account-blob/conf/01-requests-rate-status.yaml
@@ -1,0 +1,27 @@
+module: "Azure Storage Account on Blob"
+name: "requests error rate"
+filtering: "filter('resource_type', 'Microsoft.Storage/storageAccounts') and filter('primary_aggregation_type', 'true') and filter('apiname', '*Blob')"
+aggregation: ".sum(by=['azure_resource_name', 'azure_resource_group_name', 'azure_region'])"
+transformation: ".min(over='15m')"
+value_unit: "%"
+disabled: true
+signals:
+  rate_success:
+    metric: "Transactions"
+    filter: "filter('responsetype', 'Success')"
+    rollup: rate
+  rate_failed:
+    metric: "Transactions"
+    rollup: rate
+    filter: "filter('responsetype', 'ClientOtherError')"
+  signal:
+    formula:
+      (rate_failed/(rate_success+rate_failed)).scale(100).fill(0)
+rules:
+  critical:
+    threshold: 90
+    comparator: ">"
+  major:
+    threshold: 80
+    comparator: ">"
+    dependency: critical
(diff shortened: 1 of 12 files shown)


modules/integration_azure-storage-account-blob/conf/01-requests-rate-status.yaml
```diff
@@ -0,0 +1,27 @@
...
+disabled: true
+signals:
+  rate_success:
+    metric: "Transactions"
+    filter: "filter('responsetype', 'Success')"
+    rollup: rate
+  rate_failed:
+    metric: "Transactions"
+    rollup: rate
+    filter: "not filter('responsetype', 'Success')"
```
Shr3ps, line 16: hum, filter on Failed?

modules/integration_azure-storage-account-blob/conf/02-latency-e2e.yaml
```diff
@@ -0,0 +1,22 @@
...
+signals:
+  latency:
+    metric: "SuccessE2ELatency"
+    rollup: average
+  signal:
+    formula:
+      latency.scale(0.001) # Scale to s unit
+rules:
+  critical:
+    threshold: 2
```
Shr3ps, line 17: 2s of latency by default? don't you think it's low by default?

but not very important, we can override this value with the terraform variable.

modules/integration_azure-storage-account-blob/conf/readme.yaml
```diff
@@ -0,0 +1,3 @@
+documentations:
```
Shr3ps, line 1: A real link here would be great :)

Shr3ps (changes requested):

modules/integration_azure-storage-account-blob/common-modules.tf
```diff
@@ -0,0 +1,7 @@
+module "filter-tags" {
```
xp-1000, line 1: should be named `modules.tf`

modules/integration_azure-storage-account-blob/conf/readme.yaml
```diff
@@ -0,0 +1,6 @@
...
+  - name: Azure Monitor metrics
+    url: 'https://docs.microsoft.com/en-us/azure/azure-monitor/platform/metrics-supported'
+source_doc: >-
+  We are using metrics from the
+  [Microsoft.Storage/storageAccounts](https://docs.microsoft.com/en-us/azure/azure-monitor/platform/metrics-supported#microsoftclassicstoragestorageaccountsblobservices) namespace.
```
xp-1000, line 6: I would add a note about disabled detectors (why disabled by default and when to use them)

xp-1000 (changes requested):
just to minor comments

xp-1000 (approved):
LGTM thanks

Shr3ps (approved):

PR #246, opened 2021-02-17
AZ-434 Add function errors instrumentation detector

jmapro:
Add a detector for specific metrics generated by the signalfx languages wrapper (C#, Java, NodeJs, Python)

### DIFF
--- a/modules/integration_azure-functions/README.md
+++ b/modules/integration_azure-functions/README.md
@@ -7,6 +7,7 @@
 - [How to use this module?](#how-to-use-this-module)
 - [What are the available detectors in this module?](#what-are-the-available-detectors-in-this-module)
 - [How to collect required metrics?](#how-to-collect-required-metrics)
+  - [Serverless wrapper](#serverless-wrapper)
   - [Metrics](#metrics)
 - [Related documentation](#related-documentation)
 
@@ -57,7 +58,7 @@ Note the following parameters:
 
 These 3 parameters alongs with all variables defined in [common-variables.tf](common-variables.tf) are common to all 
 [modules](../) in this repository. Other variables, specific to this module, are available in 
-[variables.tf](variables.tf).
+[variables.tf](variables.tf) and [variables-gen.tf](variables-gen.tf).
 In general, the default configuration "works" but all of these Terraform 
 [variables](https://www.terraform.io/docs/configuration/variables.html) make it possible to 
 customize the detectors behavior to better fit your needs.
@@ -79,6 +80,7 @@ This module creates the following SignalFx detectors which could contain one or 
 |Azure Functions HTTP 5xx error rate|X|X|-|-|-|
 |Azure Functions connections count|X|X|-|-|-|
 |Azure Functions thread count|X|X|-|-|-|
+|Azure Functions wrapper errors|X|X|-|-|-|
 
 ## How to collect required metrics?
 
@@ -87,7 +89,19 @@ the [Azure integration](https://docs.signalfx.com/en/latest/integrations/azure-i
 with this Terraform [module](https://github.com/claranet/terraform-signalfx-integrations/tree/master/cloud/azure).
 
 
+### Serverless wrapper
 
+The detector "wrapper errors" is based on signalfx's wrapper lib which instrument your Azure function to send metrics to SignalFX.
+Metrics available with theses wrappers are :
+* azure.function.errors
+* azure.function.duration
+* azure.function.invocations
+
+The wrapper is available for following languages:
+* [C#](https://github.com/signalfx/azure-function-csharp)
+* [Java](https://github.com/signalfx/azure-function-java)
+* [NodeJs](https://github.com/signalfx/azure-function-nodejs) 
+* [Python](https://github.com/claranet/signalfx-azure-function-python)
 
 ### Metrics
 
@@ -95,6 +109,8 @@ with this Terraform [module](https://github.com/claranet/terraform-signalfx-inte
 Here is the list of required metrics for detectors in this module.
 
 * `AppConnections`
+* `azure.function.errors`
+* `azure.function.invocations`
 * `FunctionExecutionCount`
 * `Http5xx`
 * `Threads`
@@ -107,3 +123,4 @@ Here is the list of required metrics for detectors in this module.
 * [Terraform SignalFx provider](https://registry.terraform.io/providers/splunk-terraform/signalfx/latest/docs)
 * [Terraform SignalFx detector](https://registry.terraform.io/providers/splunk-terraform/signalfx/latest/docs/resources/detector)
 * [Azure Monitor metrics](https://docs.microsoft.com/en-us/azure/azure-monitor/platform/metrics-supported)
+* [SignalFx Azure Functions](https://docs.signalfx.com/en/latest/integrations/azure-functions.html)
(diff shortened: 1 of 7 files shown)


modules/integration_azure-functions/README.md
```diff
@@ -95,15 +97,19 @@ with this Terraform [module](https://github.com/claranet/terraform-signalfx-inte
...
 * `AppConnections`
+* `azure.function.errors`
 * `FunctionExecutionCount`
 * `Http5xx`
 * `Threads`
 
 
+## Notes
 
+If you want to enable azure.function.errors detector, you will need to instrusment your function. The wrapper for C#, Java and NodeJs is provided by SignalFX. But for Python, you will have to use [signalfx-azure-function-python](https://github.com/claranet/signalfx-azure-function-python)
```
xp-1000, line 108: may be can you list available repos for other languages too?

modules/integration_azure-functions/conf/01-function-errors.yaml
```diff
@@ -0,0 +1,16 @@
+
+module: functions
+name: "Azure Function errors"
```
xp-1000, line 3:
```suggestion
module: Azure Functions
name: wrapper errors
```

modules/integration_azure-functions/conf/01-function-errors.yaml
```diff
@@ -0,0 +1,16 @@
...
+  signal:
+    metric: azure.function.errors
+    extrapolation: 0
+rules:
+  critical:
+    threshold: 1
+    comparator: ">="
+filtering: "filter('is_Azure_Function', 'true')"
+aggregation: ".mean(by=['azure_function_name', 'subscription_id']"
+disabled: true
```
xp-1000, line 15: in fact it is not bad to keep this enabled by default.

if there is not any instrumented function using the wrapper so this detector will never get data so it is safe.

while you already inform the user of this dependency it seems fine for me.

modules/integration_azure-functions/conf/readme.yaml
```diff
@@ -1,3 +1,7 @@
 documentations:
   - name: Azure Monitor metrics
     url: 'https://docs.microsoft.com/en-us/azure/azure-monitor/platform/metrics-supported'
+  - name: SignalFx Azure Functions
+    url: 'https://docs.signalfx.com/en/latest/integrations/azure-functions.html'
+notes: |
+  If you want to enable azure.function.errors detector, you will need to instrusment your function. The wrapper for C#, Java and NodeJs is provided by SignalFX. But for Python, you will have to use [signalfx-azure-function-python](https://github.com/claranet/signalfx-azure-function-python)
```
xp-1000, line 7: may be can you use `source_doc` to make this even more visible like:

```
source_doc: |
  ### Serverless wrapper

  The detector xx is based on signalfx serverless wrapper lib which allow .. available on ..
```

xp-1000 (changes requested):
Cool, thanks

just some minor comments and suggestions

modules/integration_azure-functions/README.md
jmapro: Now listed in source_doc

modules/integration_azure-functions/conf/01-function-errors.yaml
jmapro: Applied.

modules/integration_azure-functions/conf/readme.yaml
jmapro: Applied.

modules/integration_azure-functions/conf/01-function-errors.yaml
jmapro: Key removed, now the detector is activated by default

modules/integration_azure-functions/conf/01-function-errors.yaml
```diff
@@ -0,0 +1,15 @@
...
+module: Azure Functions
+name: "wrapper errors"
+id: errors
+signals:
+  signal:
+    metric: azure.function.errors
+    extrapolation: 0
+rules:
+  critical:
+    threshold: 0
```
BzSpi, line 11: Instead having a static threshold, we should compare the numbers of errors with the number of invocations.

ie having at least 10% of errors for 30miutes (to discuss)

BzSpi (changes requested):

modules/integration_azure-functions/variables-gen.tf
```diff
@@ -0,0 +1,44 @@
...
+variable "errors_notifications" {
+  description = "Notification recipients list per severity overridden for errors detector"
+  type        = map(list(string))
+  default     = {}
+}
+
+variable "errors_aggregation_function" {
+  description = "Aggregation function and group by for errors detector (i.e. \".mean(by=['host'])\")"
+  type        = string
+  default     = ".mean(by=['azure_function_name', 'subscription_id']"
```
xp-1000, line 12:
```suggestion
  default     = ".mean(by=['azure_function_name', 'subscription_id'])"
```

xp-1000 (changes requested):

modules/integration_azure-functions/conf/01-function-errors.yaml
jmapro: it depend the use case. For example a function that need to send data periodically, in case of function errors you will want to know immediately that your data were not sent to take corrective actions.

But if your function is a backend for a webapp, triggering an alert on each error would be to much.

modules/integration_azure-functions/variables-gen.tf
jmapro: Fixed.

modules/integration_azure-functions/conf/01-function-errors.yaml
jmapro: Good idea. 10% of errors for 30minutes as a default value seems good. This is overridable if needed.

jmapro: Done by using the same idea as on aws lambda detectors

modules/integration_azure-functions/conf/01-function-errors.yaml
```diff
@@ -0,0 +1,28 @@
+
+module: Azure Functions
+name: "wrapper errors"
+id: errors
+signals:
+  errors:
+    metric: azure.function.errors
+    extrapolation: "last_value"
+    rollup: "average"
```
xp-1000, line 9: I would use `sum` in case of concurrent datapoints on same MTS

modules/integration_azure-functions/conf/01-function-errors.yaml
```diff
@@ -0,0 +1,28 @@
...
+  invocations:
+    metric: azure.function.invocations
+    extrapolation: "last_value"
+    rollup: "average"
```
xp-1000, line 13: I would use `sum` in case of concurrent datapoints on same MTS

xp-1000 (approved):
LGTM, we need to merge https://github.com/claranet/terraform-signalfx-detectors/pull/248 first

Shr3ps (approved):

modules/integration_azure-functions/conf/01-function-errors.yaml
jmapro: You're right, fixed in last commit.

modules/integration_azure-functions/conf/01-function-errors.yaml
jmapro: fixed in last commit.

PR #253, opened 2021-02-28
Add runbooks and tips on APIM detectors

Davidh-claranet:
Unlike others Azure detectors, the APIM detectors don't have runbook and tip variable.
I added it.
David (SG cell)


BzSpi (approved):

xp-1000 (approved):
thanks, it seems good at glance, I don't know why the CI did not run ^^

PR #254, opened 2021-03-05
Remove fill from pod_phase_status

ignacio-rivas:
cc @xp-1000


xp-1000 (approved):
Approved but please, just for the tracking could you explain in this PR description why you request this change, thanks

ignacio-rivas:
I requested this modification because the fill(2) cause the alerts to be autoresolved if the detector don't receive data for a minute. After the autoresolve, the detector create a duplicated alert for each pod.

By removing the fill(2), we gain on visibility too because we can see when the pod is created and destroyed.

PR #256, opened 2021-03-11
add kubernetes workloads count module

nicolasvion:


modules/smart-agent_kubernetes-workloads-count/conf/readme.yaml
```diff
@@ -0,0 +1,22 @@
+documentations:
+  - name: Kubernetes Deployment
+    url: 'https://kubernetes.io/docs/concepts/workloads/controllers/deployment/'
+source_doc: >-
+  We are using metrics from the
```
xp-1000, line 5: I think you can remove this sentence (or link to the "kubernetes-common" readme which should contain everything needed).

in addition to linking the right monitor in related doc (see comment above) this should be enough

modules/smart-agent_kubernetes-workloads-count/conf/readme.yaml
```diff
@@ -0,0 +1,22 @@
+documentations:
```
xp-1000, line 1: what we are interested here is the monitor used to collect metrics

```suggestion
documentations:
  - name: Smart Agent monitor
    url: 'https://docs.signalfx.com/en/latest/integrations/agent/monitors/kubernetes-cluster.html'
```

modules/smart-agent_kubernetes-workloads-count/conf/readme.yaml
```diff
@@ -0,0 +1,22 @@
...
+  - name: Kubernetes Deployment
+    url: 'https://kubernetes.io/docs/concepts/workloads/controllers/deployment/'
+source_doc: >-
+  We are using metrics from the
+  [SignalFX Support for Kubernetes](https://docs.signalfx.com/en/latest/integrations/agent/agent-k8s-install-helm.html)
+notes: |
+  ### Kubernetes Deployment Metrics
```
xp-1000, line 8:
```suggestion
  ### Kubernetes workloads count

```

modules/smart-agent_kubernetes-workloads-count/conf/readme.yaml
```diff
@@ -0,0 +1,22 @@
...
+  The default threshold values are:
+    * critical > 20
+    * major > 15 and < 20
+  If those values are not relevants, please feel free to set your own threshold.
```
xp-1000, line 12: What I looking for is an explanation of the purpose of the detector instead of default thresholds which is a redundant information available in variables values (from the registry).

```suggestion
This detector allows to set a custom limit on the total count of workloads configured on a Kubernetes cluster. This represents the sum of deployments, replicasets, statefulsets and daemonsets desired.

It can allow you set a capacity limit and be warned when it has been reached.
```

modules/smart-agent_kubernetes-workloads-count/variables-gen.tf
```diff
@@ -0,0 +1,62 @@
...
+variable "workloads_count_disabled_major" {
+  description = "Disable major alerting rule for workloads_count detector"
+  type        = bool
+  default     = null
+}
+
+variable "workloads_count_threshold_critical" {
+  description = "Critical threshold for workloads_count detector in entries"
+  type        = number
+  default     = 20
```
xp-1000, line 54: may be it could be safer to not set a default, this will force the user to define an appropriate value given that is is custom in essence.

modules/smart-agent_kubernetes-workloads-count/conf/01-kubernetes-workloads-count.yaml
```diff
@@ -0,0 +1,24 @@
...
+signals:
+  kubernetes_deployment_desired:
+    metric: "kubernetes.deployment.desired"
+  kubernetes_daemon_set_desired:
+    metric: "kubernetes.daemon_set.desired_scheduled"
+  kubernetes_replication_controller_desired:
+    metric: "kubernetes.replication_controller.desired"
+  signal:
+    formula:
+      (kubernetes_deployment_desired+kubernetes_daemon_set_desired+kubernetes_replication_controller_desired)
```
xp-1000, line 16: what about replicatset and statefulset ?

modules/smart-agent_kubernetes-workloads-count/conf/01-kubernetes-workloads-count.yaml
```diff
@@ -0,0 +1,24 @@
+module: "Kubernetes"
+name: "Workloads Count"
+filtering: "not filter('kubernetes_namespace','ara') and not filter('kubernetes_namespace','bastions') and not filter('kubernetes_namespace','gitlab-runner') and not filter('kubernetes_namespace','logging') and not filter('kubernetes_namespace','monitoring') and not filter('kubernetes_namespace','ingress-nginx') and not filter('kubernetes_namespace','kube-system') and not filter('kubernetes_namespace','kubernetes-replicator')"
+aggregation: ".sum(by=['kubernetes_name','kubernetes_namespace']).count()"
+transformation: ".min(over='30m')"
+value_unit: "entries"
```
xp-1000, line 6: I doubt "entries" is correct. "records" is probably what you look for but I am not sure it is useful to set a unit here a simple "count".

modules/smart-agent_kubernetes-workloads-count/conf/01-kubernetes-workloads-count.yaml
```diff
@@ -0,0 +1,24 @@
+module: "Kubernetes"
+name: "Workloads Count"
+filtering: "not filter('kubernetes_namespace','ara') and not filter('kubernetes_namespace','bastions') and not filter('kubernetes_namespace','gitlab-runner') and not filter('kubernetes_namespace','logging') and not filter('kubernetes_namespace','monitoring') and not filter('kubernetes_namespace','ingress-nginx') and not filter('kubernetes_namespace','kube-system') and not filter('kubernetes_namespace','kubernetes-replicator')"
```
xp-1000, line 3: this comment is optional and mainly for the record.

sadly we cannot do better using the yaml/generator config.. but we could remove the yaml conf to commit only the terraform code as "handmade" detector.

From this way you could use a terraform variable for namespaces exclusion, this will allow users to easily change or add namespaces to the default ones without to fork / PR in future.

xp-1000 (changes requested):

modules/smart-agent_kubernetes-workloads-count/conf/01-kubernetes-workloads-count.yaml
```diff
@@ -0,0 +1,26 @@
...
+    metric: "kubernetes.replication_controller.desired"
+  kubernetes_replica_set_desired:
+    metric: "kubernetes.replica_set.desired"
+  kubernetes_statefulset_desired:
+    metric: "kubernetes.stateful_set.desired"
+  signal:
+    formula:
+      (kubernetes_deployment_desired+kubernetes_daemon_set_desired+kubernetes_replication_controller_desired+kubernetes_replica_set_desired+kubernetes_statefulset_desired)
+rules:
+  critical:
```
xp-1000, line 22: if you agree with me this detector is not severe enough to be critical (see https://github.com/claranet/terraform-signalfx-detectors/wiki/Notifications-binding#severity) I would suggest to lower severity (warning / minor instead of major / critical seems good may be?)

xp-1000 (approved):
one last remark but it is good for me

modules/smart-agent_kubernetes-workloads-count/conf/01-kubernetes-workloads-count.yaml
```diff
@@ -0,0 +1,26 @@
...
+    metric: "kubernetes.replication_controller.desired"
+  kubernetes_replica_set_desired:
+    metric: "kubernetes.replica_set.desired"
+  kubernetes_statefulset_desired:
+    metric: "kubernetes.stateful_set.desired"
+  signal:
+    formula:
+      (kubernetes_deployment_desired+kubernetes_daemon_set_desired+kubernetes_replication_controller_desired+kubernetes_replica_set_desired+kubernetes_statefulset_desired)
+rules:
+  warning:
```
xp-1000, line 22: rules are reversed.

warning must depends on minor

xp-1000 (changes requested):

modules/smart-agent_kubernetes-workloads-count/conf/01-kubernetes-workloads-count.yaml
nicolasvion: indeed

xp-1000 (approved):

PR #257, opened 2021-03-18
init efs module

nicolasvion:
Some infos:
- Even if AWS Limits are high about this service, we want to keep the possibility to advice our customers by detecting problems.
- Used space and IOPs Read / Write are absolute values, it doesn't seems to have percentage interest.
- AWS Tags seems missing for this service, we have replace them from filtering by the FileSystemId var to keep distinction possibilities between env.
- Formula for IOPs are based on Cloudwatch's one


modules/integration_aws-efs/conf/03-efs-iopsread.yaml
```diff
@@ -0,0 +1,21 @@
...
+signals:
+  data_read_io_bytes:
+    metric: "DataReadIOBytes"
+    filter: "filter('stat', 'mean')"
+  meta_data_io_bytes:
+    metric: "MetadataIOBytes"
+    filter: "filter('stat', 'mean')"
+  signal:
+    formula:
+      (meta_data_io_bytes/data_read_io_bytes).scale(100)
```
xp-1000, line 15: I don't understand this one. I don't know efs but reading the cloudwatch metrics should we not monitor `MeteredIOBytes/PermittedThroughput*100` instead.

if yes, the "write" version of this detector could be deleted. If you want may be it can be replaced by the `BurstCreditBalance` which could useful to monitor.

xp-1000: this doc is awesome and give all necessary formulas to monitor throughput related metrics https://docs.aws.amazon.com/efs/latest/ug/monitoring-metric-math.html

xp-1000: I make some changes, could you check if it is ok for you @nicolasvion ?

modules/integration_aws-efs/conf/06-burst-cedit-balance.yaml
```diff
@@ -0,0 +1,14 @@
+module: "AWS EFS"
+name: "Burst Credit Balance"
+filtering: "filter('namespace', 'AWS/EFS')"
+value_unit: "credits"
+transformation: ".mean(over='5m')"
+tip: "See https://docs.aws.amazon.com/efs/latest/ug/performance.html#bursting"
+signals:
+  signal:
+    metric: "BurstCreditBalance"
+    filter: "filter('stat', 'min')"
```
nicolasvion, line 10: Doesn't seems to have min values.
Do you mean : filter: "filter('stat', 'mean')" ?

xp-1000: replace by `lower`

nicolasvion:
LGTM

PR #258, opened 2021-03-18
aws-vpn: use specific filtering because tags are not fetched

xp-1000:


xp-1000:
now the aws tags for vpn are right fetched by signalfx aws integration so this PR is useless in addition to be late !

PR #259, opened 2021-03-18
kubernetes_volumes: filter out empty volume_type

xp-1000:
in some environment it is possible to have volume metrics without the `volume_type` dimension value which are not relevant for alerting.


nicolasvion (approved):
LGTM

PR #262, opened 2021-03-19
Integration azure event hub

nicolasvion:
- Fixes based on a basic customer's case
- Microsoft.EventHub/clusters => Microsoft.EventHub/namespaces
- Default warning values at 5% (0 seems not relevant)


Shr3ps:
Can you split this in 2 PRs? maybe one for the real fixes on namespace, another for the temporary one for the tags bug.

modules/integration_azure-event-hub/README.md
```diff
@@ -95,7 +98,19 @@ Here is the list of required metrics for detectors in this module.
 * `ThrottledRequests`
 
 
+## Notes
 
+### About filtering
```
xp-1000, line 103: I don't think you should add this note until you change the default filtering.

here is an example of default filtering which does not support the default tagging convention: https://github.com/claranet/terraform-signalfx-detectors/pull/258/files

of course I let you decide what kind of filter you prefer to use to replace missing custom dimensions.

xp-1000:
without answer from the author I convert his PR to draft

PR #263, opened 2021-03-19
AZ-471: Create Azure Firewall detectors

jmapro:


modules/integration_azure_firewall/conf/00-heartbeat.yaml
```diff
@@ -0,0 +1,13 @@
+## Example
+module: firewall
+name: heartbeat
+
+transformation: false
```
xp-1000, line 5: except if you really want to make this impossible to customize I would set it to true to provide a variable to user to override default behavior

modules/integration_azure_firewall/common-modules.tf
```diff
@@ -0,0 +1 @@
+../../common/module/modules.tf
```
xp-1000, line 1: you should remove this link which works for smart_agent module only and cp one from one existing azure module (I should document this better)

modules/integration_azure_firewall/conf/00-heartbeat.yaml
```diff
@@ -0,0 +1,13 @@
+## Example
+module: firewall
```
xp-1000, line 2:
```suggestion
module: Azure Firewall
```

modules/integration_azure_firewall/conf/01-snat.yaml
```diff
@@ -0,0 +1,20 @@
+## Example
+module: firewall
```
xp-1000, line 2:
```suggestion
module: Azure Firewall
```

modules/integration_azure_firewall/conf/02-throughput.yaml
```diff
@@ -0,0 +1,27 @@
+## Example
+module: firewall
```
xp-1000, line 2:
```suggestion
module: Azure Firewall
```

modules/integration_azure_firewall/conf/03-health-status.yaml
```diff
@@ -0,0 +1,27 @@
+module: firewall
```
xp-1000, line 1:
```suggestion
module: Azure Firewall
```

modules/integration_azure_firewall/conf/readme.yaml
```diff
@@ -0,0 +1,3 @@
+documentations:
```
xp-1000, line 1:
```suggestion
documentations:
  - name: Azure Monitor metrics
    url: 'https://docs.microsoft.com/en-us/azure/azure-monitor/platform/metrics-supported'
```

modules/integration_azure_firewall/conf/01-snat.yaml
```diff
@@ -0,0 +1,20 @@
...
+name: "SNAT Port Utilization"
+
+transformation: false
```
xp-1000, line 5: except if you really want to make this impossible to customize I would set it to true to provide a variable to user to override default behavior

modules/integration_azure_firewall/conf/02-throughput.yaml
```diff
@@ -0,0 +1,27 @@
...
+    lasting_duration: "900s"
+  critical:
+    threshold: 3355443200
+    comparator: ">"
+    dependency: minor
+    lasting_duration: "900s"
+  major:
+    threshold: 4026531840
+    comparator: ">"
+    dependency: critical
```
xp-1000, line 27: it seems strange to use lasting for others rules but notthis one

modules/integration_azure_firewall/conf/02-throughput.yaml
```diff
@@ -0,0 +1,27 @@
...
+    metric: "Throughput"
+rules:
+  minor:
+    threshold: 335544320
+    comparator: ">"
+    lasting_duration: "900s"
+  critical:
+    threshold: 3355443200
+    comparator: ">"
+    dependency: minor
```
xp-1000, line 22: minor depends on major, major on critical, critical none

modules/integration_azure_firewall/conf/03-health-status.yaml
```diff
@@ -0,0 +1,27 @@
...
+name: "Health state"
+
+transformation: false
+aggregation: false
+filtering: "filter('resource_type', 'Microsoft.Network/azureFirewalls') and filter('primary_aggregation_type', 'true')"
+tip: "https://docs.microsoft.com/en-us/azure/firewall/logs-and-metrics#metrics"
```
xp-1000, line 7: I would use a full message to explain why this doc is useful (if you provide a link which is enough itself so may be `runbook_url` is better suited)

modules/integration_azure_firewall/conf/03-health-status.yaml
```diff
@@ -0,0 +1,27 @@
...
+    threshold: 50
+    comparator: "=="
+    lasting_duration: "180s"
+    lasting_at_least: "== 1.0"
+
+  major:
+    threshold: 0
+    comparator: "=="
+    lasting_duration: "180s"
+    lasting_at_least: "== 1.0"
```
xp-1000, line 26: define it to `== 1.0` results in string type which is ignored by template.

Moreover, define it as `1.0` is the same that do not define it at all.

I suggesst to remove this (same above)

xp-1000 (changes requested):

modules/integration_azure_firewall/conf/02-throughput.yaml
jmapro: It's voluntary, This is a hard limit and I want an immediate alert if reached.

modules/integration_azure_firewall/conf/01-snat.yaml
```diff
@@ -0,0 +1,21 @@
+## Example
+module: Azure firewall
+name: "SNAT Port Utilization"
+
+transformation: true
```
xp-1000, line 5: it seems very dangerous for me to not set a little timeframe on cloud metrics which are sparsed by nature and have a big interval

modules/integration_azure_firewall/conf/02-throughput.yaml
```diff
@@ -0,0 +1,27 @@
...
+  signal:
+    metric: "Throughput"
+rules:
+  critical:
+    threshold: 4026531840
+    comparator: ">"
+  major:
+    threshold: 3355443200
+    comparator: ">"
+    lasting_duration: "900s"
```
xp-1000, line 21: specific thresholds values and lasting for only one alert should be documented as notes.

it could be good to also add the unit. speaking of that, you can optionally use a `scale()` function to convert to a unit more human readable (MB/Gb).

modules/integration_azure_firewall/conf/03-health-status.yaml
```diff
@@ -0,0 +1,26 @@
...
+
+filtering: "filter('resource_type', 'Microsoft.Network/azureFirewalls') and filter('primary_aggregation_type', 'true')"
+tip: "Firewall Health state is explained here: https://docs.microsoft.com/en-us/azure/firewall/logs-and-metrics#metrics"
+
+
+signals:
+  signal:
+    metric: "FirewallHealth"
+
+# https://docs.microsoft.com/en-us/azure/firewall/logs-and-metrics#metrics
```
xp-1000, line 15: this link would be better in `related doc` readme section

xp-1000 (changes requested):

xp-1000 (changes requested):
module path must be `integration_azure-firewall` and not `integration_azure_firewall` as explained in the doc `source_name`.

modules/integration_azure-firewall/conf/03-health-status.yaml
```diff
@@ -0,0 +1,25 @@
...
+  critical:
+    threshold: 50
+    comparator: "=="
+    lasting_duration: "180s"
+
+  major:
+    threshold: 0
+    comparator: "=="
+    lasting_duration: "180s"
+    dependency: critical
```
xp-1000, line 25: you don't need dependency with `==` so either delete this or use `>`/`<` tests types

xp-1000 (changes requested):

xp-1000 (changes requested):
I see you:
- use a lasting function
- do not use a transformation

I think I understand why you do that (you want different delay for alerting for each severity rule).

I would recommend to avoid using lasting as much as possible when you define a rule dependency because it can lead to double timeframe before to trigger alert.

Here you have:
- one detector using `==` type of comparison, so you can keep lasting because you do not need dependency
- another one with `<` / `>` type of comparison and I would recommend to use a common timeframe with transformation instead of separated lasting by rule. 

I understand it is not ideal because you seem to want a different delay per rule but if you can determine an acceptable common delay for all rules, trust me it will provide a more reliable alerting experience.

If you really really want to keep lasting for dependent rules, for now my only suggestions are:
- remove dependency (it will lead to duplicated alerts if you keep different severity rules)
- do not use generator and commit the tf file manually to be able to override the `off` condition of the `when()` function for alert by a lower timeframe than the trigger one to mitigate the risk of being alerted too late.

modules/integration_azure-firewall/conf/02-throughput.yaml
```diff
@@ -0,0 +1,25 @@
...
+signals:
+  throughput:
+    metric: "Throughput"
+  signal:
+    formula: "throughput.scale(0.000000953674316)"
+rules:
+  critical:
+    threshold: 25600
+    comparator: ">"
+    lasting_duration: "900s"
```
xp-1000, line 21: a `min(over='15m')` transformation will basically do the same but you will can use dependency between rule without problem.

I left this comment for azure reviewers ;)

xp-1000 (approved):
good thanks

modules/integration_azure-firewall/conf/02-throughput.yaml
```diff
@@ -0,0 +1,25 @@
...
+  major:
+    threshold: 2560
```
BzSpi, line 23: What's the point of triggering an alert on this threshold ? 
https://docs.microsoft.com/en-us/azure/firewall/firewall-faq#how-can-i-increase-my-firewall-throughput

modules/integration_azure-firewall/conf/03-health-status.yaml
```diff
@@ -0,0 +1,24 @@
...
+
+
+signals:
+  signal:
+    metric: "FirewallHealth"
+
+rules:
+  critical:
+    threshold: 50
+    comparator: "=="
```
BzSpi, line 18: Strict equality seems very hazardous when evaluating multiple points.

BzSpi (changes requested):

modules/integration_azure-firewall/conf/02-throughput.yaml
jmapro: I think we should be alerted when the firewall need to scale up. It may be normal or not. May be a major alert is not the best level ?

modules/integration_azure-firewall/conf/03-health-status.yaml
jmapro: This metric send only 3 values: 
* 100
* 50
* 0
This is why I choose a strict equality

modules/integration_azure-firewall/conf/02-throughput.yaml
BzSpi: My 2 cents:
* critical: 29 Gbps
* Major: 27 Gbps
* Minor: 25Gbps
* Warning: 3Gbps 

Firewall scaling do not need any run operation

modules/integration_azure-firewall/conf/03-health-status.yaml
BzSpi: If your value is 50 during 2 minutes and 0 during 1 minute, your evaluation window is over and no alert has been triggered.

jmapro: What do you think if I use a transformation with `.min(over='5m')` ?
I know it is not ideal because it will not warn use on SNAT Exhaustion flapping, but I don't see how to check that.

jmapro: Ok, I used lasting instead of transformation.

Shr3ps (approved):
Good for me on the "code form", try to well test it on a customer before merge :)

PR #264, opened 2021-03-23
improve cassandra detectors

Gati0:
Currently, the aggregation is done on the delta and not on the latency itself. 
This can lead to weird results we think.
This commit fixes that.

### DIFF
--- a/modules/smart-agent_cassandra/detectors-cassandra.tf
+++ b/modules/smart-agent_cassandra/detectors-cassandra.tf
@@ -106,9 +106,9 @@ resource "signalfx_detector" "read_real_time_latency" {
   teams                   = try(coalescelist(var.teams, var.authorized_writer_teams), null)
 
   program_text = <<-EOF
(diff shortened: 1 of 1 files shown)


BzSpi (approved):

Shr3ps (approved):
LGTM

xp-1000:
I would like to have more information about these weird results before to merge

xp-1000:
see related issue here: https://github.com/claranet/terraform-signalfx-detectors/issues/278

if you agree with its description the right fix here would be to create a new "separated" variable for transformation at "formula" level (in addition the current transformation variable applied for each "metric" level).

If you go to this way, I will implement this change in the generator following your naming to make this available for all cases and fully integrated in our conventions.

also this fix (no matter how implemented) should be applied to all other "latency" detectors of this module in my opinion.

xp-1000 (changes requested):
I think it is safer to use a separate and new transformation variable for final signal as explained in #278 which will need to propagate this to all existing similar scenario.

xp-1000:
without answers from author to my question I convert this PR to draft

xp-1000:
I close because https://github.com/claranet/terraform-signalfx-detectors/pull/323 should bring similar improvements

PR #268, opened 2021-04-01
Detectors md

hlepesant:
When you have Raid Software disks managed by MD

PR #269, opened 2021-04-01
trigger ci from fork and support other sources than claranet

xp-1000:
The ci does not work on fork based PRs. it needs some explanations:

For now we only have `push` event as trigger for ci. so the steps are simply never run and the PR will remain in un mergeable state

To resolve this, it is simple because we just have to add `pull_request` event to triggers: https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull-request-events-for-forked-repositories

but historically I only kept `push` trigger to avoid duplicate checks (for internal PRs). now I understand it was a bad workaround so I gave a try to https://github.com/marketplace/actions/skip-duplicate-actions which seems to work fine.

Now it remains some problems which are related to the terraform workflow in ci (so it is more specific to our repo) and may me someone will have hints or ideas about that (I took out the @pdecat joker :p ).

This branch is in test for https://github.com/claranet/terraform-signalfx-detectors/pull/268

The remaining problem is the following:
1. github ci auto generates the terraform configuration using the `./scripts/stack/gen_module.sh` script 
2. as for any terraform config using module import we need to provide a source which was until now always the claranet github repository
3. but now we would like to support fork based PR, the source for terraform requires to be set on the right repository: claranet for internal PR or fork if external PR
4. of course this script as for all others are also used by humans so it needs to still work in this scenario

I need to check values of all available variables https://docs.github.com/en/actions/reference/environment-variables especially : 

- GITHUB_REF 
- GITHUB_HEAD_REF 
- GITHUB_BASE_REF

If we can get the info it is a fork from them we could probably implement a hack in the script.

Another possible way would be to use github workflow rule with something like:

`if: github.event.pull_request.head.repo.full_name != 'claranet/terraform-signalfx-detectors'`

to set the variable and use it in the script.

Any opinion ? or may be do you already known similar problems?

### DIFF
--- a/.github/workflows/doc.yml
+++ b/.github/workflows/doc.yml
@@ -2,14 +2,27 @@
 
 name: Documentation
 
-on:
-  push:
-#    paths:
-#    - '**.md'
-#    - '.github/workflows/doc.yml'
+on: [push, pull_request]
 
 jobs:
+  pre_job:
+    # continue-on-error: true # Uncomment once integration is finished
+    runs-on: ubuntu-latest
+    # Map a step output to a job output
+    outputs:
+      should_skip: ${{ steps.skip_check.outputs.should_skip }}
+    steps:
+    - id: skip_check
+      name: Skip Duplicate Actions
+      uses: fkirc/skip-duplicate-actions@v3.4.0
+      with:
+        concurrent_skipping: 'same_content'
+        do_not_skip: '["pull_request", "workflow_dispatch", "schedule"]'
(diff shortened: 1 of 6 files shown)


xp-1000:
ok so it seems not possible to fully support PRs from forks sadly or, at least, not without huge modification of the testing workflow.

Indeed, we need a SignalFx token to push detectors to a test organization for
testing purpose and we cannot share this token to external user because of
obvious security reasons.

It is far to be the best but we basically do not run terraform checks from an external repository so all related ci steps are ignored when token is not available.

this pr also brings general improvements like using `skip-duplicate-actions`, support different github owner in `init-stack` command and trigger `pull_request` event to enable rest of CI for PRs from forks.

Poil (approved):

PR #271, opened 2021-04-09
add detectors for hpa

### LINKED ISSUE #270: [RFE] New detector hpa for module kubernetes-common
**What is the module?**
https://github.com/claranet/terraform-signalfx-detectors/tree/master/modules/smart-agent_kubernetes-common.

**Is your feature request related to a known monitoring hole? Please describe.**
HPA usage must be monitored to ensure hpa sizing.

**Describe the solution you'd like**
An usage ratio in order to detect full hpa which resizing.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

egouraud-claranet:
resolve #270

### DIFF
--- a/modules/smart-agent_kubernetes-common/conf/01-hpa-scale-capacity.yaml
+++ b/modules/smart-agent_kubernetes-common/conf/01-hpa-scale-capacity.yaml
@@ -1,0 +1,15 @@
+module: kubernetes
+name: "hpa scale exceeded capacity"
+tip: "hpa ask to scale for too long, the maximum number of desired Pods has been hit or there is a lack of resources"
+signals:
+  max:
+    metric: kubernetes.hpa.spec.max_replicas
+  desired:
+    metric: kubernetes.hpa.status.desired_replicas
+  signal:
+    formula: (desired-max)
+rules:
+  major:
+    threshold: 0
+    comparator: ">="
+    lasting_duration: "300s"
--- a/docs/severity.md
+++ b/docs/severity.md
@@ -764,6 +764,7 @@
 
 |Detector|Critical|Major|Minor|Warning|Info|
 |---|---|---|---|---|---|
+|Kubernetes hpa scale exceeded capacity|-|X|-|-|-|
 |Kubernetes node heartbeat|X|-|-|-|-|
 |Kubernetes node status|-|X|X|-|-|
 |Kubernetes pod status phase|-|X|-|-|-|
--- a/modules/smart-agent_kubernetes-common/README.md
+++ b/modules/smart-agent_kubernetes-common/README.md
@@ -64,7 +64,7 @@ Note the following parameters:
 
 These 3 parameters alongs with all variables defined in [common-variables.tf](common-variables.tf) are common to all 
 [modules](../) in this repository. Other variables, specific to this module, are available in 
-[variables.tf](variables.tf).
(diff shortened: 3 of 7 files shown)


xp-1000 (commented):
Thanks for the contribution.

I transformed your hcl code into a yaml config file to use the jinja generator, this drastically ease the review (and maintainability) and is good enough for simple scenario like this.

It should be equivalent but please tell me if I am wrong.

I have one comment about the detector formula I confess I am not sure to understand, can you review my suggestion or explain to me what it is not obvious ?

Also I think checking the `kubernetes.hpa.status.condition.able_to_scale` metric could be a good addition to your detector which will not cover all cases of scaling unavailability (maybe?)

modules/smart-agent_kubernetes-common/conf/01-hpa-scale-capacity.yaml
```diff
@@ -0,0 +1,15 @@
...
+  max:
+    metric: kubernetes.hpa.spec.max_replicas
+  current:
+    metric: kubernetes.hpa.status.current_replicas
+  signal:
+    formula: (max-current)
+rules:
+  major:
+    threshold: 0
+    comparator: "!="
```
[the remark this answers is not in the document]
egouraud-claranet: erf, my bad, indeed my mistake was reverse thinking 
i want to raise an alert if current replicas number is on max replicas for too long, so `==` instead of `!=` (if difference is 0 for too long)

modules/smart-agent_kubernetes-common/conf/01-hpa-scale-capacity.yaml
```diff
@@ -0,0 +1,15 @@
...
+  max:
+    metric: kubernetes.hpa.spec.max_replicas
+  current:
+    metric: kubernetes.hpa.status.current_replicas
+  signal:
+    formula: (max-current)
+rules:
+  major:
+    threshold: 0
+    comparator: "="
```
xp-1000, line 14: Ok this time it makes sens for me :) however I would like to suggest:

```suggestion
  desired:
    metric: kubernetes.hpa.status.desired_replicas
  signal:
    formula: (max-desired)
rules:
  major:
    threshold: 0
    comparator: ">="
```

I think using `desired` instead of `current` could provide a better result:
- raising alert sooner: as soon as the configuration is pushed (before to wait it has been applied)
- covering all cases: I am afraid if the condition could not be meet so your current will never scale and you will not have any alert whereas it is not able to scale

Notice the correct operator is `==` and not `=` but I also prefer to keep the `>=` which seems more suited here while we want to alert only if current or desired are > to max (and not the reverse).

If you agree you can apply this suggestion directly from github web.

xp-1000 (changes requested):

modules/smart-agent_kubernetes-common/conf/01-hpa-scale-capacity.yaml
egouraud-claranet: I agree with the reasonning, but i will also change the alert description, it should be "hpa is not able to scale up" (imagine a lack of compute resources), the alert is more general and can manage two kinds of issues: lack of ressources, too small hpa

my first intention was only to monitor a full hpa ;)

xp-1000 (approved):
good job!

PR #248, opened 2021-02-18
Feature/lasting

jmapro:
Closes #247 
Needed for #246

PR #274, opened 2021-04-14
generator: lasting is per rule only and mandatory like threshold

xp-1000:
the lasting variable is "mandatory" (created as soon it has been configured) and its scope is `rule` only exactly like thresholds

the disabled is little special because it is also available at `detector` and `global` scope. But this is useless to provide a per rule variable if there is only one rule given that disabling the entire detector will does the same than disabling its only rule.

When we integrated this feature from https://github.com/claranet/terraform-signalfx-detectors/pull/248 I did not this error because we did not have any detector with only one rule using lasting.

so it works until now but the proper way to implement this is define lasting in the same block we define thresholds (and not the disabled per rule one).

I add @jmapro as reviewer to let him now this change on this feature.

@egouraud-claranet you should be able to fix the ci on https://github.com/claranet/terraform-signalfx-detectors/pull/271 after rebasing on this PR.

### DIFF
--- a/modules/integration_azure-firewall/variables-gen.tf
+++ b/modules/integration_azure-firewall/variables-gen.tf
@@ -150,22 +150,10 @@ variable "throughput_disabled_critical" {
   default     = null
 }
 
-variable "throughput_lasting_duration_critical" {
-  description = "Minimum duration that conditions must be true before raising alert"
-  type        = string
-  default     = "300s"
-}
-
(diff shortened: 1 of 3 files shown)


xp-1000:
wiki updated: https://github.com/claranet/terraform-signalfx-detectors/wiki/Variables#lasting_duration

egouraud-claranet (approved):

PR #275, opened 2021-04-14
Integration Azure Event Hub fixes

Shr3ps:
fixes based on #262 (but keep only standard changes)

### DIFF
--- a/modules/integration_azure-event-hub/conf/readme.yaml
+++ b/modules/integration_azure-event-hub/conf/readme.yaml
@@ -7,8 +7,6 @@ notes: |
   Unlike other Azure services, the tags defined by user on Event Hubs are not fetched by SignalFx Azure
   integration as dimensions. This is why we cannot filter on [the azure tagging
   convention](https://github.com/claranet/terraform-signalfx-detectors/wiki/Tagging-convention#azure).
-  In this case, you could have to leverage multiple instances of this module to monitor different Event Hubs providing
-  different value for this entity_name. Else, its default value * will monitor all Event Hubs.
 
   ### About warning default values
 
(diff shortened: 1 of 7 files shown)


modules/integration_azure-event-hub/conf/readme.yaml
```diff
@@ -1,3 +1,13 @@
...
+notes: |
+  ### About filtering
+
+  Unlike other Azure services, the tags defined by user on Event Hubs are not fetched by SignalFx Azure
+  integration as dimensions. This is why we cannot filter on [the azure tagging
+  convention](https://github.com/claranet/terraform-signalfx-detectors/wiki/Tagging-convention#azure).
+
+  ### About warning default values
+
+  Basic customer's case seems to report values beetween 0 and 5% in a nominal situation.
```
xp-1000, line 13:
```suggestion
  ### About `throttled_requests` detector

  Basic customer's case seems to report values between 0 and 5% in a nominal situation so the default warning threshold has been set to `5`.
```

xp-1000 (changes requested):
just a little suggestion, if you disagree just fix the typo and we will be good.

modules/integration_azure-event-hub/conf/readme.yaml
```diff
@@ -1,3 +1,13 @@
 documentations:
   - name: Azure Monitor metrics
     url: 'https://docs.microsoft.com/en-us/azure/azure-monitor/platform/metrics-supported'
+notes: |
+  ### About filtering
```
xp-1000, line 5: I don't think you should add this note until you change the default filtering.

here is an example of default filtering which does not support the default tagging convention: https://github.com/claranet/terraform-signalfx-detectors/pull/258/files

of course I let you decide what kind of filter you prefer to use to replace missing custom dimensions.

xp-1000 (approved):

PR #279, opened 2021-04-27
Add missing curly bracket in AWS default filters

NSenaud:
It looks like a curly bracket is missing here. I am not sure what the impact is, but it may be related to some unjustified heartbeat notifications we are receiving sometimes (we are currently testing the fix, but it is to soon to draw conclusions).


xp-1000 (approved):
thanks!

PR #282, opened 2021-05-24
Add aggregation by subscription_id on GCP Pub/Sub Subscription heartbeat

ignacio-rivas:
Aggregation by subscription_id on the detector GCP Pub/Sub Subscription heartbeat for avoiding the false positives when we have a request with response_code=cancelled

### DIFF
--- a/modules/integration_gcp-pubsub-subscription/detectors-subscription.tf
+++ b/modules/integration_gcp-pubsub-subscription/detectors-subscription.tf
@@ -6,7 +6,7 @@ resource "signalfx_detector" "heartbeat" {
 
   program_text = <<-EOF
     from signalfx.detectors.not_reporting import not_reporting
-    signal = data('subscription/pull_request_count', filter=${module.filter-tags.filter_custom}).sum(by=['subscription_id'])${var.heartbeat_aggregation_function}.publish('signal')
+    signal = data('subscription/pull_request_count', filter=${module.filter-tags.filter_custom})${var.heartbeat_aggregation_function}.publish('signal')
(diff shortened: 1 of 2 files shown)


modules/integration_gcp-pubsub-subscription/detectors-subscription.tf
```diff
@@ -6,7 +6,7 @@ resource "signalfx_detector" "heartbeat" {
 
   program_text = <<-EOF
     from signalfx.detectors.not_reporting import not_reporting
-    signal = data('subscription/pull_request_count', filter=${module.filter-tags.filter_custom})${var.heartbeat_aggregation_function}.publish('signal')
+    signal = data('subscription/pull_request_count', filter=${module.filter-tags.filter_custom}).sum(by=['subscription_id'])${var.heartbeat_aggregation_function}.publish('signal')
```
xp-1000, line 9: the way you fix it is good by forcing aggregation instead of let signalfx to check every single mts.

another way would be to add a filter to exclude `response_code:cancelled` but I suppose other code could exist and lead to similar problems.

That said, I think you should define this aggregation as default for `heartbeat_aggregation_function` variable instead of hardcode it in the signalflow just before this variable.

this "double aggregation" is not obvious for the user and it seems fully relevant as default value as it allow the user to change this behavior which remain only a recommendation (as for majority of this repo content in true).

please let me know if you have a reason to hardcode this I could miss

xp-1000 (changes requested):

xp-1000 (approved):

PR #285, opened 2021-06-10
auto resolve heartbeat alerts after mts is considered as inactive

xp-1000:
when a mts does not report during 36h it is considered as inactive by Splunk but the possible triggered alerts will not be auto cleared.

it seems not relevant to keep these alerts while they have no chance to be resolved normally one day without an active mts.

Splunk just has released a feature to auto resolve alert so using it with `36h`  delay should auto clear inactive mts related alerts.

it mainly concern heartbeat detectors but it can also be useful to rare others but for non heartbeat detector it seems better to use `fill()` or extrapolation function when we meet the case (as we already do).

See official documentation https://docs.signalfx.com/en/latest/detect-alert/zombie-alerts.html

### DIFF
--- a/modules/integration_aws-alb/detectors-alb.tf
+++ b/modules/integration_aws-alb/detectors-alb.tf
@@ -7,7 +7,7 @@ resource "signalfx_detector" "heartbeat" {
   program_text = <<-EOF
     from signalfx.detectors.not_reporting import not_reporting
     signal = data('RequestCount', filter=filter('stat', 'sum') and filter('namespace', 'AWS/ApplicationELB') and ${module.filtering.signalflow})${var.heartbeat_aggregation_function}.publish('signal')
-    not_reporting.detector(stream=signal, resource_identifier=None, duration='${var.heartbeat_timeframe}').publish('CRIT')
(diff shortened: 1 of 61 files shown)


modules/integration_aws-alb/detectors-alb.tf
```diff
@@ -7,7 +7,7 @@ resource "signalfx_detector" "heartbeat" {
   program_text = <<-EOF
     from signalfx.detectors.not_reporting import not_reporting
     signal = data('RequestCount', filter=filter('stat', 'sum') and filter('namespace', 'AWS/ApplicationELB') and ${module.filter-tags.filter_custom})${var.heartbeat_aggregation_function}.publish('signal')
-    not_reporting.detector(stream=signal, resource_identifier=None, duration='${var.heartbeat_timeframe}').publish('CRIT')
+    not_reporting.detector(stream=signal, resource_identifier=None, duration='${var.heartbeat_timeframe}', auto_resolve_after='1s').publish('CRIT')
```
Shr3ps, line 10: Hum, what about a constant in `common-locals.tf`? Would be easier for you to update every detectors in the future :)

Shr3ps (changes requested):

pdecat (approved):
LGTM!

modules/integration_aws-alb/detectors-alb.tf
xp-1000: done

Shr3ps (approved):

PR #286, opened 2021-06-10
filter_custom now is a string and supports append mode

### LINKED ISSUE #103: Append mode for custom filter tags
Currently, custom filtering only allow to replace default convention by another one.

could be nice to allow 2 different modes of custom fitlering:

- replace: as currently it will erase default tags which should be redefined to be kept
- append: allow to add new tags to default convention which will be kept without to redefine them

xp-1000:
resolve #103 adding a way to append custom filters to the default ones instead of replace them.

But it also changes the behavior of the filter-tags module by dropping the 2 list variables "exclude"/"include" in favor of a simple filter_custom variable which is a full signalflow string as for the filter_defaults one.

this should provide more flexibility to the user to apply complex filtering and fully enjoy signalflow capability (it also simplify the code of this module).

Here are examples of the output of the filter-tags module:

- using default filtering only (follow the convention inside the repo):
```
TF_VAR_filter_defaults="filter('env', 'test') and filter('sfx_monitored', 'true')" terraform plan

Changes to Outputs:
  + filter_custom = "filter('env', 'test') and filter('sfx_monitored', 'true')"
```

- using custom filtering only (follow its own convention or no convention at all):
```
TF_VAR_filter_defaults="filter('env', 'test') and filter('sfx_monitored', 'true')" TF_VAR_filter_custom="filter('foo', 'bar')"  terraform plan

Changes to Outputs:
  + filter_custom = "filter('foo', 'bar')"
```

- using the default filtering convention but append custom common filters to it:

```
TF_VAR_filter_defaults="filter('env', 'test') and filter('sfx_monitored', 'true')" TF_VAR_filter_custom="filter('foo', 'bar')" TF_VAR_append_mode=true terraform plan

Changes to Outputs:
  + filter_custom = "(filter('env', 'test') and filter('sfx_monitored', 'true')) and (filter('foo', 'bar'))"
```

the append mode behavior simply produces: `(filter_defaults) and (filter_custom)`.

TODO: 
- [x] update all modules
- [x] update the wiki and explain the new behavior


common/module/modules.tf
```diff
@@ -1,8 +1,8 @@
 module "filter-tags" {
-  source = "github.com/claranet/terraform-signalfx-detectors.git//common/filter-tags"
+  source = "github.com/claranet/terraform-signalfx-detectors.git//common/filter-tags?ref=103_improve_filtering_and_append_mode"
```
xp-1000, line 2: reset the ref to master

xp-1000: not useful anymore because we now use local sourcing of this module

modules/integration_aws-alb/README.md
```diff
@@ -43,7 +43,7 @@ Note the following parameters:
   instance of the module.
   Its value will be added to the `prefixes` list at the start of the [detector 
   name](https://github.com/claranet/terraform-signalfx-detectors/wiki/Templating#example).
-  In general, it will also be used in `filter-tags` sub-module to apply a
+  In general, it will also be used in `filtering` internal sub-module to apply a
```
pdecat, line 46: I think this should be changed to either `apply a filter` or `apply filtering`.

pdecat (approved):

modules/integration_aws-alb/README.md
xp-1000: `apply a filter(s)` is good for you?

pdecat: Simply `apply filters` would be better.

xp-1000: done !

Shr3ps (approved):

modules/internal_filtering/variables.tf
```diff
@@ -0,0 +1,18 @@
+variable "filtering_default" {
+  description = "Filters as SignalFlow string to enforce the default filtering convention from the template"
+  type        = string
+  default     = ""
+}
+
+variable "filtering_custom" {
+  description = "Filters as SignalFlow string to override the default filtering convention from the user input"
+  type        = string
+  default     = ""
```
Shr3ps, line 10: Maybe it's not a common use-case, but shouldn't be better to set "null" by default, and make your test in output "== null"?

This would allow us to possibily set filtering_custom="" -> no filtering at all (a real empty string is outputed)

xp-1000: done!

PR #287, opened 2021-06-17
add back filter-tags module for retro compatibility

xp-1000:
add back the "old" filter-tags module to keep all versions <= 1.6.0 working before the 1.7.0 which will use the new module with `filtering_custom` variable and relative path source.

### DIFF
--- a/common/filter-tags/README.md
+++ b/common/filter-tags/README.md
@@ -1,0 +1,96 @@
+# FILTER TAGS SignalFx
+
+__DEPRECATION WARNING__: this module is kept only for retro compatibility of versions prior to `v1.7.0`.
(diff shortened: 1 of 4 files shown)


common/filter-tags/README.md
```diff
@@ -0,0 +1,96 @@
+# FILTER TAGS SignalFx
+
+__DEPRECATION WARNING__: this module is kept only for retro compatibility of versions prior to `v1.7.0`.
```
pdecat, line 3:
```suggestion
__DEPRECATION WARNING__: this module is kept only for retro compatibility of versions prior to `v1.7.0` that referenced it without specifying a tag.
```

pdecat (approved):
LGTM!

PR #295, opened 2021-06-25
Add pm2 detectors

jphilaine:


modules/smart-agent_pm2/conf/00-heartbeat.yaml
```diff
@@ -0,0 +1,14 @@
+## Example
```
xp-1000, line 1: you can remove this ;)

modules/smart-agent_pm2/conf/01-up.yaml
```diff
@@ -0,0 +1,15 @@
+module: pm2
+name: "up application"
```
xp-1000, line 2:
```suggestion
name: "application status"
```
?

modules/smart-agent_pm2/conf/readme.yaml
```diff
@@ -0,0 +1,3 @@
+documentations:
```
xp-1000, line 1: could be good to add some information how to setup pm2 metrics with an example of agent configuration to fetch them.

also it seems useful to explain why `(not filter('name', 'pm2-metrics'))` is required

xp-1000 (commented):
some suggestions here

PR #297, opened 2021-07-02
Decrease cache_hit_rate criticity

BenjyOC:
I updated the cache_hit_rate detector criticity.


xp-1000 (changes requested):
I would say to use minor/warning combination because this remains a performance related issue https://github.com/claranet/terraform-signalfx-detectors/wiki/Notifications-binding#minor

BenjyOC:
@xp-1000 Why not. A low hit cache ratio can generate poor http application performance but not because of varnish service. However, it can be useful to alert customer about it with minor/warning severity. I'll update the code.
Thx !

xp-1000:
Hello @BenjyOC, any news ?

I would like to release a new version, just want to check if I can include this

PR #299, opened 2021-07-05
Fix Postgresql resource type name

maxpoullain:
I updated the resource type name for the PostgreSQL detectors.
It was broken and not working anymore.

### DIFF
--- a/modules/integration_azure-postgresql/detectors-postgresql.tf
+++ b/modules/integration_azure-postgresql/detectors-postgresql.tf
@@ -7,7 +7,7 @@ resource "signalfx_detector" "heartbeat" {
 
   program_text = <<-EOF
         from signalfx.detectors.not_reporting import not_reporting
-        base_filter = filter('resource_type', 'Microsoft.DBForPostgreSQL/servers') and filter('primary_aggregation_type', 'true')
+        base_filter = filter('resource_type', 'Microsoft.DB*orPostgreSQL/servers') and filter('primary_aggregation_type', 'true')
         signal = data('network_bytes_ingress', filter=base_filter and ${module.filtering.signalflow})${var.heartbeat_aggregation_function}.publish('signal')
         not_reporting.detector(stream=signal, resource_identifier=None, duration='${var.heartbeat_timeframe}', auto_resolve_after='${local.heartbeat_auto_resolve_after}').publish('CRIT')
     EOF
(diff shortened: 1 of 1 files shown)


xp-1000 (approved):

Shr3ps:
Since when this is broken? On which tenant did you test?
Official doc says: https://docs.microsoft.com/en-us/azure/azure-monitor/essentials/metrics-supported#microsoftdbforpostgresqlservers

maxpoullain:
> Since when this is broken? On which tenant did you test?
> Official doc says: https://docs.microsoft.com/en-us/azure/azure-monitor/essentials/metrics-supported#microsoftdbforpostgresqlservers

No idea since when this is broken. It's my first time implementing it.

What I'm sure though, it's that I was correctly getting metrics with the Azure integrations but no data was found on the detectors themselves.
Upon further inspection of the Signalflow, this is the difference I found: 
Metrics are getting scrapped from the resource type name `Microsoft.DBForPostgreSQL/servers` from the Azure integration but on the PostgreSQL detector, the resource type name is `Microsoft.DBforPostgreSQL/servers`, with no data getting reported.
I tried my fix and it's working fine now.

maxpoullain:
Updated with `Microsoft.DB*orPostgreSQL/servers` to match both PostgreSQL resource type name.

xp-1000:
hello @maxpoullain the readme order is not right, may be you updated it manually instead of using the makefile ?

xp-1000:
@Shr3ps is it ok for you now ?

modules/integration_azure-postgresql/detectors-postgresql.tf
```diff
@@ -33,7 +33,7 @@ resource "signalfx_detector" "cpu_usage" {
   tags                    = compact(concat(local.common_tags, local.tags, var.extra_tags))
 
   program_text = <<-EOF
-        base_filter = filter('resource_type', 'Microsoft.DBforPostgreSQL/servers') and filter('primary_aggregation_type', 'true')
+        base_filter = filter('resource_type', 'Microsoft.DB?orPostgreSQL/servers') and filter('primary_aggregation_type', 'true')
```
BzSpi, line 36:
```suggestion
        base_filter = filter('resource_type', 'Microsoft.DBForPostgreSQL/servers', 'Microsoft.DBforPostgreSQL/servers') and filter('primary_aggregation_type', 'true')
```
More readable in my opinion

Shr3ps (approved):

PR #302, opened 2021-07-08
Logtash detectors module.

chungktran:


chungktran:
@xp-1000 Could you tell me what I need to do to pass the documentation tests? Thanks.

xp-1000:
Hello @chungktran 

thanks for the contribution!

The following block contains optional hints, jump below if you want answers to your question.

---

Did you consider to use the [yaml generator](https://github.com/claranet/terraform-signalfx-detectors/tree/master/scripts/templates) to build your detectors?

Not mandatory, just a suggestion, but it should be easier to auto generatecode and especially pass all CI checks with less effort.

When you have a module directory structure  as yours, you can update it adding new detectors by:

1. create a new `[0-9][0-9]-mydetector.yaml` file into the `conf` dir of the exiting module which will be used after to automatically generate the changes. you can copy [this file](https://github.com/claranet/terraform-signalfx-detectors/blob/master/scripts/templates/values.yaml) which is similar to an helm chart and contains all available options. you can also check existing examples like [system-common](https://github.com/claranet/terraform-signalfx-detectors/tree/master/modules/smart-agent_system-common/conf).
1. run docker and then enter in dev env with `make`
1. run `make update-module *logstash*`
1. commit and push and the CI should be happy

There is a [TLDR](https://github.com/claranet/terraform-signalfx-detectors/blob/master/docs/development.md#tldr) here to create a new module from scratch

---

first you can check[this](https://github.com/claranet/terraform-signalfx-detectors/blob/master/docs/development.md#checks) which contains a description of each CI check and what `make` command (or script) you have to run to auto update your source and pass it. 

if you still want to do all the required changes manually so, in general, it comes down to read the failed CI output and change your source to match its diff.

For example here:
- the `Documentation / modules` [check output](https://github.com/claranet/terraform-signalfx-detectors/pull/302/checks?check_run_id=3021705273) shows that it miss the `Logstash cpu usage percent` in the detectors list of the readme (and other diff too).
- for `Documentation / severity`, this is because you have to update the global list of detectors in https://github.com/claranet/terraform-signalfx-detectors/blob/master/docs/severity.md. This is basically a copy of the list of detectors available in the readme of your module.
- for `Detectors / outputs` [check output](https://github.com/claranet/terraform-signalfx-detectors/pull/302/checks?check_run_id=3021705766) we can see the terraform output for `cpu_usage_percent` is right defined but not ordered in the file. Indeed, the auto generation script used in the CI (and also optionally from contributor with the makefile) sort them alphabetically and so it see a git diff.. I am sorry for that may be checking that the output is preset would be enough ..

xp-1000:
I edited the readme yaml file and run `make update-module *logstash*` on your fork, the CI should pass now

chungktran:
Thanks @xp-1000. I'll definitely look into yaml generator. I didn't know about it that's why I did it manually.

modules/smart-agent_logstash/variables.tf
```diff
@@ -0,0 +1,375 @@
...
+variable "cpu_usage_percent_transformation_function" {
+  description = "Transformation function for cpu_usage_percent detector (i.e. \".mean(over='5m')\")"
+  type        = string
+  default     = ".mean(over='5m')"
+}
+
+variable "cpu_usage_percent_threshold_critical" {
+  description = "Critical threshold for cpu_usage_percent detector"
+  type        = number
+  default     = 100
```
xp-1000, line 355: Even if you right used `>=` comparison I am afraid `100` could not be always reached. In doubt I would set `99` or `99.8` if you still want to keep a very high value.

modules/smart-agent_logstash/variables.tf
```diff
@@ -0,0 +1,375 @@
...
+}
+
+variable "cpu_usage_percent_threshold_critical_lasting" {
```
xp-1000, line 358: the lasting replaces threshold here. this is a separate setting which is right on `rule` scope as indicated here https://github.com/claranet/terraform-signalfx-detectors/wiki/Variables#lasting_duration so we still keep the severity in name so:

```
variable "cpu_usage_percent_lasting_duration_critical" {
```

same for all other lasting variable

modules/smart-agent_logstash/variables.tf
```diff
@@ -0,0 +1,375 @@
...
+variable "events_out_high_transformation_function" {
+  description = "Transformation function for events_out_high detector (i.e. \".mean(over='5m')\")"
+  type        = string
+  default     = ".mean(over='5m')"
+}
+
+variable "events_out_high_threshold_critical" {
+  description = "Critical threshold for events_out_high detector"
+  type        = number
+  default     = 30000
```
xp-1000, line 210: may be some thresholds should not have default defined, this will force the user to set its own given that it is variable depending on each environment

xp-1000 (changes requested):
It looks good overall :clap: 

I added some suggestions.

I also have a global advice about `lasting` function usage and I would recommend to drop it and simply replace the transformation function to `min` (with `>` comparator) and `max` (with `<` comparator). This will behaves exactly the same than a lasting delay here (wait 5m of consecutive datapoints reached the threshold before to raise alert) but we more reliable result.

Indeed, even if it is a powerful and awesome function it can lead to undesired delay when used with multiple severity rules as here. I will try to explain simply with my limited english :p 

The only known problematic situation where it can show problem is when:
- used with multi dependent severity rules like here (the major severity exclude the critical test to avoid duplicated alerts when value is above critical threshold)
- the data varies between 2 thresholds but does not stay long time enough in the same threshold. for example, `critical > value > major` during 2m then `value > critical > major` during 2 mn and again and again ..

in this case, the delay will be the double of your lasting for 2 dependent rules so here it can takes up to 10m before to raise the first alert instead of 5m configured as lasting.

For a short lasting this is not a big deal but if we suppose a user configure `1h` as lasting, the double value could become problematic.

Using a `min` / `max` transformation does not suffer of this limitation so it is, in general, better to use it even if it will smooth the signal displayed in chart compared to the underlying metric. Or, it is also possible to keep lasting but defining only one rule so there is not problem anymore.

chungktran:
@xp-1000 Thanks for all your suggestions. I will look into making the suggested changes.

xp-1000:
hello @chungktran

I convert this PR as draft.
It will be easy for me to know when review/merge when you will make it ready.

chungktran:
@xp-1000 Sorry that I haven't had time to get back into this with my workload. I'll definitely revisit this soon enough. Thanks for your patience.

xp-1000:
> @xp-1000 Sorry that I haven't had time to get back into this with my workload. I'll definitely revisit this soon enough. Thanks for your patience.

no problem, take you time ;)

chungktran:
Closing this PR for now until I make the recommended changes.

PR #306, opened 2021-08-05
smart-agent_systemd-services: init

tchernomax:
new module


modules/smart-agent_systemd-services/README.md
```diff
@@ -0,0 +1,112 @@
+# SYSTEMD-SERVICES SignalFx detectors
```
xp-1000, line 1: may be `systemd` would be enough and shorter for the module name ?

modules/smart-agent_systemd-services/conf/readme.yaml
```diff
@@ -0,0 +1,5 @@
+documentations:
+  - name: Smart Agent monitor
+    url: 'https://docs.signalfx.com/en/latest/integrations/agent/monitors/collectd-systemd.html'
+
+source_doc:
```
xp-1000, line 5: I would add a configuration example of the monitor here 
you check an example here : https://github.com/claranet/terraform-signalfx-detectors/tree/master/modules/smart-agent_nginx-ingress

modules/smart-agent_systemd-services/detectors-systemd-services.tf
```diff
@@ -0,0 +1,24 @@
+resource "signalfx_detector" "systemd_services" {
```
xp-1000, line 1: given that the detector is very basic I would have used the yaml generator to ease the creation and the review.

See https://github.com/claranet/terraform-signalfx-detectors/blob/master/docs/development.md#bootstrap-a-new-module for more information and https://github.com/claranet/terraform-signalfx-detectors/blob/master/modules/smart-agent_system-common/conf/01-cpu.yaml for an example

xp-1000 (changes requested):
Overall seems good, just some remarks

modules/smart-agent_systemd-services/README.md
tchernomax: I disagree, today systemd manage A LOT of things.
Beside .service there is .timer, .path, .device, .mount, …

With this module we can only monitor services and I think it should be explicit in the module name.

xp-1000: I thought maybe this module could grow up in future to embed more than services but ok no problem just a suggestion

xp-1000 (approved):
Good job thanks!

PR #309, opened 2021-08-06
smart-agent_systemd-services: better comparator

tchernomax:


modules/smart-agent_systemd-services/conf/00-systemd_services.yaml
```diff
@@ -12,4 +12,4 @@ signals:
 rules:
```
xp-1000, line 12: I missed this in the previous PR and this is the scope of the current one but please could you increment the file id like `01-substate.yaml` ? (I think using `systemd` is repetitive in a systemd module but the only important thing is to keep `00` for heatbeat ;) )

xp-1000 (changes requested):
a little remark in comment and also I would recommend to:

- set aggregation to true
- remove the `id` definition

sorry all these remarks are out of scope

modules/smart-agent_systemd-services/conf/00-systemd_services.yaml
tchernomax: done

tchernomax:
@xp-1000 all remarks have been done.

xp-1000 (approved):
LGTM

PR #311, opened 2021-08-06
detector: `when(X) and when(Y)` → `when(X and Y)`

tchernomax:
`when` predicate support logical operators :
https://dev.splunk.com/observability/docs/signalflow/functions/when_function/#Syntax

I don't know if it's better to separate the when or merge them… just food for thought.


xp-1000 (commented):
It is a good idea and that is true for now there is no good reason to separate.

But in true the lasting function should be applied separately so 2 when are necessary, this is just not done yet ...

tchernomax:
close and replaced by https://github.com/claranet/terraform-signalfx-detectors/pull/314

PR #314, opened 2021-08-09
generator: dependency is now negative instead of reversed and lasting always available

tchernomax:
# fix

generator dependency severity bug

# and

remove "negative_comparator" variable:
```
when(X negative_comparator Y) → not when(X dependency_comparator Y)
```


xp-1000 (approved):
Awesome!

I will try to refect this change to all the manual tf code also before to merge this one.

thanks

tchernomax:
# added

lasting handling

PR #316, opened 2021-08-13
negative dependency logic instead of reverse

xp-1000:
finish the work of https://github.com/claranet/terraform-signalfx-detectors/pull/314 by:

- applying the same change (negative instead of reverse dependency condition) on manually written detectors using a batch command
- manage some exceptions of this command by converting them to use the generator and fully and automatically enjoy best practices easily: `integration-aws_alb`, `integration-aws_elb`, `integration-aws_apigateway`, `integration-gcp_load-balancing` and `smart-agent_nginx-ingress`.


tchernomax (approved):
LGTM (I didn't review all the changes, but the concept is OK)

PR #317, opened 2021-08-16
generator: add append_condition field to rule

xp-1000:
add ability to append an optional condition to the default thresholds based ones from the generator


tchernomax (approved):
LGTM

PR #318, opened 2021-08-16
harmonize lasting usage

xp-1000:
following work from https://github.com/claranet/terraform-signalfx-detectors/pull/314 this PR tries to harmonize lasting usage using variable `lasting_duration` at rule level as described in our doc https://github.com/claranet/terraform-signalfx-detectors/wiki/Variables.

- it does not touch to existing manually written detectors which do not provide lasting (function and variable)
- it converts all existing detectors using the deprecated `lasting_duration_seconds` variable to use the generator and automatically enjoy all new features and conventions
- it manually change to `lasting_duration` with proper negation for multi rules detectors using `timer` undocumented variable


tchernomax (approved):
LGTM (I didn't review all the changes, but the concept is OK)

PR #326, opened 2021-09-15
Fix Postgresql resource type name

maxpoullain:
Hello,

Following up on this MR https://github.com/claranet/terraform-signalfx-detectors/pull/299.

The regex with `Microsoft.DB?orPostgreSQL/servers` doesn't work on my project but `Microsoft.DB*orPostgreSQL/servers` is working correctly.

I don't think we were able to test with the `Microsoft.DB?orPostgreSQL/servers` before it was merged.


xp-1000:
> I don't think we were able to test with the Microsoft.DB?orPostgreSQL/servers before it was merged.

If PR is not ready for merge so it should be a draft ;)

xp-1000 (approved):
When I merged I did not see the last change from `*` to `?` which is subsequent to your last comment https://github.com/claranet/terraform-signalfx-detectors/pull/299#issuecomment-874571208 and my approve

maxpoullain:
> If PR is not ready for merge so it should be a draft ;)

You are right, that was my bad.  
This version has been tested on my side though.

PR #328, opened 2021-09-28
optimize heartbeat delay

xp-1000:
Heartbeat detectors are a bit special and they depend on lot of factors especially due to the exclusion of not running cloud instances to prevent false alert on downscaled instances.
So even for metric with short interval, like agents metrics by default to 10 they depend on the metadata sync interval from cloud integration:

- the detector job interval: 10s
- the metrics collection interval: 10s
- the mts resolution: 10s
- cloud metadata synchronization: 15m

Previously we integrated a `max_delay` of 15m in https://github.com/claranet/terraform-signalfx-detectors/pull/80 to ensure metadata synchronization BEFORE to send any alert (which could be false because metadata are not up to date and a terminated instance could still be considered as alive falsly).

It works fine and now we never got new false alert but in worst case it can so delay the alert to 15m (in addition to the heartbeat timeframe of 20m).

We cannot shorter the max_delay because it must match the cloud metadata synchronization interval of 15m else we will get back undesired false positive. 

A request feature exists on SignalFx side (FEED-3354) to be able to shorter this job but waiting for this, the only way is to shorter the heartbeat timeframe instead.
From 20m to 10m the delay will be: 10m (heartbeat timeframe) + 15m (max_delay) = 25m max

It seems to be a good default value to speed up heartbeat alert, keeping it safe and users can customize this heartbeat timeframe as they want case by case


rsicart:
For smartagent modules  `modules/smart-agent_*`, can't we set a lower `max_delay`?
Or we need to wait for instance status (from cloud integrations) anyways?

xp-1000:
> For smartagent modules `modules/smart-agent_*`, can't we set a lower `max_delay`?
> Or we need to wait for instance status (from cloud integrations) anyways?

you always can set lower max_delay but you will probably get false alert because you are not sure the metadata sync has been done before alert.

PR #329, opened 2021-10-01
Draft: Create Azure Express Route detectors

jmapro:
Create detectors for Azure Express Route


xp-1000:
@BzSpi or @Shr3ps could you take a look on this PR given that it is related to azure ?

PR #341, opened 2021-11-04
add systemd-timers module

NSenaud:
New detectors to monitor Systemd Timers (and replace crons with better observability).

### DIFF
--- a/modules/smart-agent_systemd-timers/conf/00-heartbeat.yaml
+++ b/modules/smart-agent_systemd-timers/conf/00-heartbeat.yaml
@@ -1,0 +1,12 @@
+module: systemd-timers
+name: heartbeat
+
+transformation: false
+aggregation: true
+disabled: true
+
+signals:
+  signal:
+    metric: "gauge.substate.failed"
+rules:
+  minor:
--- a/modules/smart-agent_systemd-timers/conf/01-execution-delay.yaml
+++ b/modules/smart-agent_systemd-timers/conf/01-execution-delay.yaml
@@ -1,0 +1,18 @@
+module: systemd-timers
+name: "execution delay"
+aggregation: true
+transformation: true
+disabled: true
+signals:
+  A:
+    metric: gauge.active_state.active
+  B:
+    metric: gauge.active_state.activating
+  signal:
+    formula: "(A+B)"
+rules:
+  major:
+    threshold: 1
+    comparator: "<"
+    description: "is too long"
+    lasting_duration: 1d
--- a/modules/smart-agent_systemd-timers/conf/readme.yaml
+++ b/modules/smart-agent_systemd-timers/conf/readme.yaml
@@ -1,0 +1,37 @@
+documentations:
+  - name: Smart Agent monitor collectd/systemd
+    url: https://docs.signalfx.com/en/latest/integrations/agent/monitors/collectd-systemd.html
+
+source_doc: |
+  This module checks Systemd services launched by Systemd timers. The difference with [smart-agent_systemd-services](https://github.com/claranet/terraform-signalfx-detectors/tree/master/modules/smart-agent_systemd-services) is that services launched by timers are not supposed to be always running, and instead can be stopped most of the time.
+
+  Detectors are designed to check that the job did not failed (enabled by default), the service have not been removed (disabled by default), and the job have been launched during the past day (disabled by default, you can override the default delay with the `execution_delay_lasting_duration_major` variable).
+
+  For instance, to enable all the optional detectors and be alerted if the service haven't been launched during the past week, you can use the following options:
+  ```hcl
+  execution_delay_disabled = false
+  heartbeat_disabled       = false
+
+  execution_delay_lasting_duration_major = "1w"
+  ```
+
+  ### Examples
+
+  You can configure the systemd collector in the agent the following way to report only the state of the `my_daily_job` and `my_weekly_job` services and the metrics needed by the detector:
+  ```yaml
+      - type: collectd/systemd
+        services:
+        - my_daily_job
+        - my_weekly_job
+        sendActiveState: true
+        datapointsToExclude:
+        - metricNames:
+          - '*'
+          - '!gauge.active_state.activating'
+          - '!gauge.active_state.active'
+          - '!gauge.substate.failed'
+        extraMetrics:
+        - gauge.substate.failed
+        - gauge.active_state.active
+        - gauge.active_state.activating
+  ```
--- a/docs/severity.md
+++ b/docs/severity.md
@@ -96,6 +96,7 @@
 - [supervisor](#supervisor)
 - [system-common](#system-common)
 - [systemd-services](#systemd-services)
+- [systemd-timers](#systemd-timers)
 - [tomcat](#tomcat)
 - [varnish](#varnish)
 - [zookeeper](#zookeeper)
@@ -1030,6 +1031,15 @@
 |Detector|Critical|Major|Minor|Warning|Info|
 |---|---|---|---|---|---|
 |Systemd-services aliveness|X|-|-|-|-|
+
+
+## systemd-timers
+
+|Detector|Critical|Major|Minor|Warning|Info|
+|---|---|---|---|---|---|
+|Systemd-timers heartbeat|-|-|X|-|-|
+|Systemd-timers execution delay|-|X|-|-|-|
+|Systemd-timers last execution state|-|X|-|-|-|
 
 
 ## tomcat
--- a/modules/smart-agent_systemd-timers/README.md
+++ b/modules/smart-agent_systemd-timers/README.md
@@ -1,0 +1,149 @@
+# SYSTEMD-TIMERS SignalFx detectors
+
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+:link: **Contents**
+
+- [How to use this module?](#how-to-use-this-module)
+- [What are the available detectors in this module?](#what-are-the-available-detectors-in-this-module)
+- [How to collect required metrics?](#how-to-collect-required-metrics)
+  - [Examples](#examples)
+  - [Metrics](#metrics)
+- [Related documentation](#related-documentation)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## How to use this module?
+
+This directory defines a [Terraform](https://www.terraform.io/) 
+[module](https://www.terraform.io/docs/modules/usage.html) you can use in your
+existing [stack](https://github.com/claranet/terraform-signalfx-detectors/wiki/Getting-started#stack) by adding a 
+`module` configuration and setting its `source` parameter to URL of this folder:
+
+```hcl
+module "signalfx-detectors-smart-agent-systemd-timers" {
+  source = "github.com/claranet/terraform-signalfx-detectors.git//modules/smart-agent_systemd-timers?ref={revision}"
+
+  environment   = var.environment
+  notifications = local.notifications
+}
+```
+
+Note the following parameters:
+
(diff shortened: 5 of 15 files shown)


modules/smart-agent_systemd-timers/conf/00-heartbeat.yaml
```diff
@@ -0,0 +1,11 @@
...
+name: heartbeat
+
+transformation: false
+aggregation: true
+
+signals:
+  signal:
+    metric: "gauge.substate.failed"
+rules:
+  minor:
```
xp-1000, line 11: this is not usual to use such a low severity. in general we use critical one because it is related to availability according to https://github.com/claranet/terraform-signalfx-detectors/wiki/Notifications-binding#critical. can you explain the choice ?

modules/smart-agent_systemd-timers/conf/01-timer-has-not-run-for.yaml
```diff
@@ -0,0 +1,15 @@
+module: systemd
+name: "timer has not run for"
```
xp-1000, line 2: not a big fan of this name. a detector's name should ideally:
- describe the detector goal (ok here)
- be a complete sentence (ko here)
keep in mind that the name will be what is displayed on signalfx ui and in our doc like: https://github.com/claranet/terraform-signalfx-detectors/blob/feat-systemd-timers-detectors/docs/severity.md#systemd-timers

what about:
```suggestion
name: "execution delay"
```

This will create a detector with the name:
> [env] Systemd-timers execution delay
with alerts like:
> [env] Systemd-timers execution delay is too high (XX > YY)

If you want you can add a `description` attribute to the `rule` to customize the message append to detector's name like:
```hcl
rules:
  major:
    threshold: 1
    comparator: "<"
    description: "is too long"
    lasting_duration: 30m
```

modules/smart-agent_systemd-timers/conf/01-timer-has-not-run-for.yaml
```diff
@@ -0,0 +1,15 @@
+module: systemd
```
xp-1000, line 1:
```suggestion
module: systemd-timers
```

modules/smart-agent_systemd-timers/conf/02-timer-has-failed.yaml
```diff
@@ -0,0 +1,13 @@
+module: systemd
```
xp-1000, line 1:
```suggestion
module: systemd-timers
```

modules/smart-agent_systemd-timers/conf/00-heartbeat.yaml
```diff
@@ -0,0 +1,11 @@
+module: systemd-timers
```
xp-1000, line 1: I am not sure if it is a good idea to have an hearbeat in this case, can you confirm the metric reporting is regular and relevant?

modules/smart-agent_systemd-timers/conf/01-timer-has-not-run-for.yaml
```diff
@@ -0,0 +1,15 @@
...
+name: "timer has not run for"
+transformation: false
```
xp-1000, line 3: I think it is safe to give the user the ability to set a transformation if he wants

```suggestion
transformation: true
```

modules/smart-agent_systemd-timers/conf/01-timer-has-not-run-for.yaml
```diff
@@ -0,0 +1,15 @@
...
+    metric: gauge.active_state.active
+  B:
+    metric: gauge.active_state.activating
+  signal:
+    formula: "(A+B)"
+rules:
+  major:
+    threshold: 1
+    comparator: "<"
+    lasting_duration: 30m
```
xp-1000, line 15: Seems ok for a first version but we should probably think for something more generic

modules/smart-agent_systemd-timers/conf/02-timer-has-failed.yaml
```diff
@@ -0,0 +1,13 @@
...
+name: "timer has failed"
```
xp-1000, line 2:
```suggestion
name: "last execution state"
```

with a rule description set to `is failed` ?

modules/smart-agent_systemd-timers/conf/02-timer-has-failed.yaml
```diff
@@ -0,0 +1,13 @@
...
+
+transformation: true
+aggregation: false
```
xp-1000, line 5:
```suggestion
aggregation: true
```

modules/smart-agent_systemd-timers/conf/readme.yaml
```diff
@@ -0,0 +1,8 @@
+documentations:
+  - name: Smart Agent monitor collectd/systemd
+    url: https://docs.signalfx.com/en/latest/integrations/agent/monitors/collectd-systemd.html
+
+source_doc: |
+  This module checks Systemd services launched by Systemd timers. The difference with [smart-agent_systemd-services](https://github.com/claranet/terraform-signalfx-detectors/tree/master/modules/smart-agent_systemd-services) is that services launched by timers are not supposed to be always running, and instead can be stopped most of the time.
+
+  Detectors are designed to check that the job succeeded, the service have not been removed, and the job have been lanched during the past 30 minutes by default (you can override this behavior with the `timer_has_not_run_for_lasting_duration_major` variable).
```
xp-1000, line 8:
```suggestion
  Detectors are designed to check that the job succeeded, the service have not been removed, and the job have been launched during the past 30 minutes by default (you can override this behavior with the `timer_has_not_run_for_lasting_duration_major` variable).
```

xp-1000 (changes requested):
Thanks for the contribution.

I left some comments and suggestions

modules/smart-agent_systemd-timers/conf/00-heartbeat.yaml
NSenaud: IMHO crons/timers are usually used for batch jobs, which in most cases are not critical. I don't see a use case where a cron/timer create an availability issue platform-wide. But if it better fits with the default values used elsewhere, I still can set a critical level by default, and can send the notification on BH alerting channel if my use case.

modules/smart-agent_systemd-timers/conf/00-heartbeat.yaml
NSenaud: It is regular, relevant it depends: the alert will be triggered if the service is removed. I think it could be useful, especially on instances which aren't replaced on a regular basis such as a bastion.

modules/smart-agent_systemd-timers/conf/01-timer-has-not-run-for.yaml
NSenaud: I included your suggestion :checkered_flag:

modules/smart-agent_systemd-timers/conf/01-timer-has-not-run-for.yaml
NSenaud: As discussed in direct message, it is not easy to make it generic, so it is disabled by default and the default alerting is set to 1 day instead of 30 minutes.

modules/smart-agent_systemd-timers/conf/00-heartbeat.yaml
xp-1000: I confess I am not sure but perhaps you are right and there are less "critical" cronjob than "minor" ones.
it is up to you so ;)

modules/smart-agent_systemd-timers/conf/readme.yaml
```diff
@@ -0,0 +1,8 @@
+documentations:
+  - name: Smart Agent monitor collectd/systemd
+    url: https://docs.signalfx.com/en/latest/integrations/agent/monitors/collectd-systemd.html
+
+source_doc: |
+  This module checks Systemd services launched by Systemd timers. The difference with [smart-agent_systemd-services](https://github.com/claranet/terraform-signalfx-detectors/tree/master/modules/smart-agent_systemd-services) is that services launched by timers are not supposed to be always running, and instead can be stopped most of the time.
+
+  Detectors are designed to check that the job succeeded, the service have not been removed, and the job have been launched during the past 30 minutes by default (you can override this behavior with the `timer_has_not_run_for_lasting_duration_major` variable).
```
xp-1000, line 8:
```suggestion
  Detectors are designed to check that the job succeeded, the service have not been removed, and the job have been launched during the past 30 minutes by default (you can override this behavior with the `execution_delay_lasting_duration_major` variable).
```

xp-1000 (changes requested):
Can you add a note about the disabled detectors in "notes" section of the readme ?

Last question and we should be good: did you test this module in WITH the `smart-agent_systemd-services` deployed ? I am afraid there could be undesired conflicts or border cases between them. Indeed, as soon as you gather metrics for both: a "timer" service and a "normal" service, detectors of both module could apply different rules on same services no?

modules/smart-agent_systemd-timers/conf/01-execution-delay.yaml
```diff
@@ -0,0 +1,17 @@
+module: systemd-timers
+name: "execution delay"
+transformation: true
```
xp-1000, line 3:
```suggestion
aggregation: true
transformation: true
```

modules/smart-agent_systemd-timers/conf/readme.yaml
NSenaud: Should be better now. Also, I can confirm both detectors can be used together, of course you will most likely not monitor the same services with both of them, but if you use `filtering_custom` this is not a problem.

NSenaud: @xp-1000 I cleaned up history, I let you confirm everything looks good to you now :)

modules/smart-agent_systemd-timers/conf/readme.yaml
```diff
@@ -0,0 +1,35 @@
...
+  Detectors are designed to check that the job did not failed (enabled by default), the service have not been removed (disabled by default), and the job have been launched during the past day (disabled by default, you can override the default delay with the `execution_delay_lasting_duration_major` variable).
+
+  For instance, to enable all the optional detectors and be alerted if the service haven't been launched during the past week, you can use the following options:
+  ```tf
+  execution_delay_disabled = false
+  heartbeat_disabled       = false
+
+  execution_delay_lasting_duration_major = "1w"
+  ```
+
```
xp-1000, line 17:
```suggestion

### Examples

```

modules/smart-agent_systemd-timers/conf/readme.yaml
```diff
@@ -0,0 +1,35 @@
...
+  - name: Smart Agent monitor collectd/systemd
+    url: https://docs.signalfx.com/en/latest/integrations/agent/monitors/collectd-systemd.html
+
+source_doc: |
+  This module checks Systemd services launched by Systemd timers. The difference with [smart-agent_systemd-services](https://github.com/claranet/terraform-signalfx-detectors/tree/master/modules/smart-agent_systemd-services) is that services launched by timers are not supposed to be always running, and instead can be stopped most of the time.
+
+  Detectors are designed to check that the job did not failed (enabled by default), the service have not been removed (disabled by default), and the job have been launched during the past day (disabled by default, you can override the default delay with the `execution_delay_lasting_duration_major` variable).
+
+  For instance, to enable all the optional detectors and be alerted if the service haven't been launched during the past week, you can use the following options:
+  ```tf
```
xp-1000, line 11:
```suggestion
  ```hcl
```

xp-1000 (approved):

modules/smart-agent_systemd-timers/conf/readme.yaml
xp-1000: hi @NSenaud it looks good to me, just 2 minor comments.

PR #347, opened 2021-11-15
clean tmp files when lint fails

xp-1000:
from the input of @Poil we should clean tmp files when lint fails and do not reach normal end.

### DIFF
--- a/scripts/module/lint.sh
+++ b/scripts/module/lint.sh
@@ -4,11 +4,11 @@ set -ue -o pipefail
 out_file="tmp-outputs.tf"
 
 # Clean when exit
-err() {
+clean() {
     rm -f ${out_file}
 }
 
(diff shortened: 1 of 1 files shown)


scripts/module/lint.sh
```diff
@@ -30,4 +39,4 @@ done
 
 echo "Lint module ${TARGET}"
 eval "$TFLINT_CMD"
-rm tmp-outputs.tf
+rm ${out_file}
```
Poil, line 42: `rm -f` to avoid error in case of no common-variables ?

xp-1000: good catch

xp-1000: should be better now

Shr3ps (approved):

PR #348, opened 2021-11-15
Exclude squashfs from filesystem detector

Poil:
Note : It's better to exclude it on agent side


xp-1000 (changes requested):
good, please could you add a note in the readme to inform the user it excludes this and perhaps to recommend to do that at collection level too ?

modules/smart-agent_system-common/conf/03-space.yaml
```diff
@@ -6,6 +6,7 @@ value_unit: "%"
 signals:
   signal:
     metric: disk.utilization
+    filter: "not filter('fs_type', 'squashfs')"
```
xp-1000, line 9:
```suggestion
    filter: "(not filter('fs_type', 'squashfs'))"
```

xp-1000 (changes requested):

xp-1000:
you must remove auto format on markdown. this must be done on the template or it will lead to a diff to the next generation

PR #356, opened 2021-12-14
Fix deadlettered message queue detector

Davidh-claranet:
the "entityname" field has been renamed to "EntityName". (The detctor was broken)
The major threshold has been changed to 0.

### DIFF
--- a/modules/integration_azure-service-bus/conf/01-deadlettered-messages.yaml
+++ b/modules/integration_azure-service-bus/conf/01-deadlettered-messages.yaml
@@ -15,7 +15,7 @@ rules:
     threshold: 10
     comparator: ">"
   major:
-    threshold: 5
+    threshold: 0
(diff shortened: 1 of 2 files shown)


xp-1000:
@Shr3ps @BzSpi can you take a look please?

BzSpi (approved):

xp-1000:
@Davidh-claranet @BzSpi @Shr3ps  the fix for dimension name should not be applied to other detectors too ?

Davidh-claranet:
"EntityName" seems to only be used in the datafactory detector

xp-1000:
> "EntityName" seems to only be used in the servicebus detector

I spoke about other detectors in this module, it seems weird from azure to only change the dimension for one metric of service bus and not the others but I don't know azure :shrug:

PR #362, opened 2021-12-28
Azure MySQL detectors: convert to yaml configs

Shr3ps:


modules/integration_azure-mysql/conf/00-heartbeat.yaml
```diff
@@ -0,0 +1,13 @@
+module: Azure MySQL
+name: heartbeat
+
+transformation: true
+aggregation: ".mean(by=['azure_resource_id'])"
```
xp-1000, line 5: I am afraid this id could change too often and cause undesirable alerts.
using same aggregation than other detectors could make this heartbeat less sensible to user change ? perhaps there is no possible change on it or you really want to be alerted of this kind of change, just a remark.
in any case using the id is not the most understandable when you receive an alert, in this case, do not use aggregation at all should provide same alerting behavior but the user will get all information instead of the id only.

modules/integration_azure-mysql/conf/01-cpu.yaml
```diff
@@ -0,0 +1,22 @@
+module: Azure MySQL
+name: "CPU usage"
```
xp-1000, line 2: if you want you can define id like
`id: cpu`
will make variables naming shorter and "usage" is not really useful for id.
can be applied to other detectors too

modules/integration_azure-mysql/conf/05-replication-lag.yaml
```diff
@@ -0,0 +1,22 @@
+module: Azure MySQL
+name: "replication lag"
+
+transformation: true
+aggregation: ".mean(by=['azure_resource_name', 'azure_resource_group_name', 'azure_region'])"
+
+filtering: "filter('resource_type', 'Microsoft.DBforMySQL/servers') and filter('primary_aggregation_type', 'true')"
+value_unit: "seconds"
```
xp-1000, line 8:
```suggestion
value_unit: "Second"
```

xp-1000 (changes requested):
only unit `Second` related comment is mandatory, others are up to you

modules/integration_azure-mysql/conf/00-heartbeat.yaml
Shr3ps: I kept the previous detector (written manually in TF) behavior which already deployed on many customers.
That's not really the purpose of this MR to challenge past detector (just wanted to rewrite using yaml conf)

xp-1000: ok ok no problem

xp-1000 (approved):
looks good, thanks

PR #363, opened 2021-12-28
Azure MariaDB detectors module

Shr3ps:


modules/integration_azure-mariadb/conf/05-replication-lag.yaml
```diff
@@ -0,0 +1,22 @@
+module: Azure MariaDB
+name: "replication lag"
+
+transformation: true
+aggregation: ".mean(by=['azure_resource_name', 'azure_resource_group_name', 'azure_region'])"
+
+filtering: "filter('resource_type', 'Microsoft.DBforMariaDB/servers') and filter('primary_aggregation_type', 'true')"
+value_unit: "seconds"
```
xp-1000, line 8:
```suggestion
value_unit: "Second"
```

modules/integration_azure-mariadb/conf/00-heartbeat.yaml
```diff
@@ -0,0 +1,13 @@
+module: Azure MariaDB
+name: heartbeat
+
+transformation: true
+aggregation: ".mean(by=['azure_resource_id'])"
```
xp-1000, line 5: I am afraid this id could change too often and cause undesirable alerts.
using same aggregation than other detectors could make this heartbeat less sensible to user change ? perhaps there is no possible change on it or you really want to be alerted of this kind of change, just a remark.
in any case using the id is not the most understandable when you receive an alert, in this case, do not use aggregation at all should provide same alerting behavior but the user will get all information instead of the id only.

modules/integration_azure-mariadb/conf/01-cpu.yaml
```diff
@@ -0,0 +1,22 @@
+module: Azure MariaDB
+name: "CPU usage"
```
xp-1000, line 2: defining id like
`id: cpu`
will make variables naming shorter and "usage" is not really useful here.
can be applied to other detectors if you want.

xp-1000 (approved):
seems good, just some minor comments

xp-1000 (approved):
if we find other similar cases where 2 modules are almost the same we could find a way to handle that and limit the duplication. I think it is enough for now. thanks

Issue #377, opened 2022-02-10
[RFE] New module of detectors for AWS Redshift


gbloquel:
**Please describe the new component you want to monitor**
AWS Redshift

**What is the source and how to collect metrics?**
https://docs.splunk.com/Observability/gdi/get-data-in/connect/aws/aws-metrics.html#amazon-redshift

**How to monitor?**
* CPU Utilization
* Heartbeat
* Percentage disk space used

PR #378, opened 2022-02-10
fix: #377 Integrate AWS redshift

gbloquel:


modules/integration_aws-redshift/conf/00-heartbeat.yaml
```diff
@@ -0,0 +1,13 @@
+module: AWS Redshift
+name: heartbeat
+
+transformation: false
+aggregation: ".mean(by=['ClusterIdentifier', 'NodeID'])"
```
xp-1000, line 5: Should we check if all nodes respond or only the cluster itself ?
I don't have the answer I honestly do not know redshift.

modules/integration_aws-redshift/conf/readme.yaml
```diff
@@ -0,0 +1,5 @@
+documentations:
+  - name: AWS metrics for Redshift
+    url: 'https://docs.splunk.com/Observability/gdi/get-data-in/connect/aws/aws-metrics.html#amazon-redshift'
```
xp-1000, line 3: perhaps you can add the official aws doc which can lead to improvements or new detectors in future ? https://docs.aws.amazon.com/redshift/latest/mgmt/metrics-listing.html

xp-1000 (approved):
perfect job

modules/integration_aws-redshift/conf/00-heartbeat.yaml
gbloquel: You're right. I need to check the cluster only.

PR #381, opened 2022-02-14
smart-agent_system-common: rename "per_cpu_enabled" variable to "agent_per_cpu_enabled"

BzSpi:
smart-agent_system-common: rename "per_cpu_enabled" variable to "agent_per_cpu_enabled"


xp-1000 (approved):
@BzSpi @Poil do you think we should use `agent_cpu_average_enabled` instead to use the option from otel collector rather that smart agent ?

PR #382, opened 2022-02-15
Update documentation links

xp-1000:
This PR aims to fix broken links now docs.signalfx.com is dead.

It also include some related improvement like adding the splunk observability doc link in addition to signalfx smart agent monitor one.

The CI will fails to keep this PR clean for review, see https://github.com/claranet/terraform-signalfx-detectors/pull/383 for fixed state

### DIFF
--- a/modules/integration_azure-app-service-plan/conf/readme.yaml
+++ b/modules/integration_azure-app-service-plan/conf/readme.yaml
@@ -1,3 +1,3 @@
 documentations:
   - name: Azure Monitor metrics
-    url: 'https://docs.microsoft.com/en-us/azure/azure-monitor/platform/metrics-supported#microsoftwebserverfarms'
+    url: 'https://docs.microsoft.com/en-us/azure/azure-monitor/essentials/metrics-supported#microsoftwebserverfarms'
--- a/modules/integration_azure-app-service-plan/conf/readme.yaml
+++ b/modules/integration_azure-app-service-plan/conf/readme.yaml
@@ -1,3 +1,3 @@
 documentations:
   - name: Azure Monitor metrics
-    url: 'https://docs.microsoft.com/en-us/azure/azure-monitor/platform/metrics-supported#microsoftwebsitesslots'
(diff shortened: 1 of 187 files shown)


BzSpi (approved):

modules/integration_azure-app-service-plan/conf/readme.yaml
```diff
@@ -1,3 +1,3 @@
 documentations:
   - name: Azure Monitor metrics
-    url: 'https://docs.microsoft.com/en-us/azure/azure-monitor/platform/metrics-supported'
+    url: 'https://docs.microsoft.com/en-us/azure/azure-monitor/platform/metrics-supported#microsoftwebsitesslots'
```
Shr3ps, line 3:
```suggestion
    url: 'https://docs.microsoft.com/en-us/azure/azure-monitor/essentials/metrics-supported#microsoftwebserverfarms'
```

modules/integration_azure-functions/conf/readme.yaml
```diff
@@ -1,8 +1,6 @@
 documentations:
-  - name: Azure Monitor metrics
-    url: 'https://docs.microsoft.com/en-us/azure/azure-monitor/platform/metrics-supported'
```
Shr3ps, line 3: https://docs.microsoft.com/en-us/azure/azure-monitor/essentials/metrics-supported#microsoftwebsites

Shr3ps: With dimension "is Azure Function : true"

Shr3ps (changes requested):

modules/integration_azure-functions/conf/readme.yaml
xp-1000: oh good catch, I removed the doc because I only opended the `detectors-gen.tf` file and understood the metric does not come from azure integration so I supposed that the used metrics in module only come from signalfx serverless wrapper.

I missed the "manually" written detectors file which indeed use azure integrations metrics.

Shr3ps (approved):

PR #398, opened 2022-03-21
Fix integration_azure-mariadb

zfiel:


xp-1000:
Hello,

it is not the first time, we noticed the same for postgresql: https://github.com/claranet/terraform-signalfx-detectors/pull/299

Microsoft still does not manage sensitive in 2022..

As for the previous PR I recommend to use `*` wildcard for any char because I am pretty sure that the 2 patterns exist on different orgs.

Shr3ps:
> Hello,
> 
> it is not the first time, we noticed the same for postgresql: #299
> 
> Microsoft still does not manage sensitive in 2022..
> 
> As for the previous PR I recommend to use `*` wildcard for any char because I am pretty sure that the 2 patterns exist on different orgs.

👍 
Let's use again '*'

zfiel:
> Hello,
> 
> it is not the first time, we noticed the same for postgresql: #299
> 
> Microsoft still does not manage sensitive in 2022..
> 
> As for the previous PR I recommend to use `*` wildcard for any char because I am pretty sure that the 2 patterns exist on different orgs.

Done.

Shr3ps (approved):

xp-1000 (approved):

xp-1000:
@BzSpi @Shr3ps should you no do the same for azure-mysql?

PR #400, opened 2022-03-30
Support empty notifications list

tchernomax:


xp-1000 (approved):
LGTM thanks

in fact I follow https://github.com/hashicorp/terraform/issues/19898 I would like to convert the notficiations variables from map to object and define a default.

while this feature is still experimental this way is good for now

cvauvarin (approved):

PR #325, opened 2021-08-31
smart agent redis

tchernomax:
smart-agent_redis : transform to generator (+ use lasting sometimes)

PR #409, opened 2022-05-05
smart-agent_redis : transform to generator (+ use lasting sometimes)

tchernomax:
rework of https://github.com/claranet/terraform-signalfx-detectors/pull/325


scripts/templates/values.yaml
```diff
@@ -74,6 +79,11 @@ transformation: true
 #filtering: "filter('primary_aggregation_type', 'true')"
 filtering:
 
+## @param filtering_suffix - string - optional - default: ' and '
+## Determine the string to append at the end of the "filtering" param.
+#
+filtering_suffix: ' and '
```
xp-1000, line 85: while you set a default in the template could you remove this value from this yaml?

scripts/templates/detector.tf.j2
```diff
@@ -8,9 +8,12 @@
 {%- else -%}
   {%- set type = 'threshold' -%}
 {%- endif -%}
-{%- set filtering_variable = 'base_filtering' -%}
```
xp-1000, line 11: we must keep the filtering variable.

this is a way to make the signalflow more readable (splitting in different lines, the main line for signal definition is already too long ..)

and this will avoid all the unrelated changes on this PR

xp-1000 (changes requested):
Hello,

Thanks for you work, 

I like the idea but I leaved 2 comments for changes which are important to me.

scripts/templates/values.yaml
xp-1000: also if you can handle the spaces inside the template to not need to put them here when set this would be more user friendly

tchernomax: > while you set a default in the template could you remove this value from this yaml?
done

> also if you can handle the spaces inside the template to not need to put them here when set this would be more user friendly
I prefer not to, leaving the space in the variable make the template more configurable.

If I move the space to the template I think we should rename the variable (as it's no longer a suffix).
If you still think we should move the space from the variable to the template, what variable name should we use ?

xp-1000: I seen spaces as a safety feature which does not have limitation undesired impact but I can be wrong.

I am not against to force the user to put the spaces I just don't see the advantage.

Can you give me an example where it is useful ?

PR #412, opened 2022-05-10
Genericjmx: add Parallel Old GC and Concurrent Mark Sweep GC support

tchernomax:
my client use multiple types of GC…

The only one I didn't try is CMS, but I found the memorypool name here: https://gist.github.com/huynhbaoan/1bb4cd87313f56bc63cc4e19ce5c3ec7

### DIFF
--- a/modules/smart-agent_genericjmx/conf/02-gc_old.yaml
+++ b/modules/smart-agent_genericjmx/conf/02-gc_old.yaml
@@ -3,8 +3,15 @@ name: GC old gen
 
 transformation: ".min(over='5m')"
 value_unit: "%"
-#                                for G1 GC                                               for Serial GC
-filtering: (filter('plugin_instance', 'memory_pool-G1 Old Gen') or filter('plugin_instance', 'memory_pool-Tenured Gen'))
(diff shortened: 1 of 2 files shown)


xp-1000:
hello @tchernomax are you sure if a dimension is available on memory used so it is on memory max too and vice versa ?

else you could produce a wrong average (above or below 100%)

tchernomax:
> are you sure if a dimension is available on memory used so it is on memory max too and vice versa ?
> else you could produce a wrong average (above or below 100%)

Yes, I am quasi sure of it.

Every jmx "memory usage" provide 4 values: "init", "used", "commited" and "max"
https://docs.oracle.com/cd/E17802_01/j2se/j2se/1.5.0/jcp/beta1/apidiffs/java/lang/management/MemoryUsage.html

Those values are retrieved from retrieve from the same jmx beam attribute (it return a hash with the 4 values).

In the 'plugin_instance=memory_pool-XXX' dimension, the XXX is the name of the jmx beam.
As, "used" and "max" are from the same attribute of the same beam → if "used" is present with dimension 'plugin_instance=memory_pool-TOTO', "max" is present with the same dimension.

xp-1000 (approved):

PR #416, opened 2022-05-31
OpenTelemetry dimension is `type` and not `fs_type`

Poil:
Not sure if we should open a case to splunk, the dimension should not changed when migrating to otel


pdecat (approved):
LGTM

Shr3ps (approved):

PR #417, opened 2022-06-02
Change permitted.scale(1024) -> permitted.scale(60), need to be in se…

gburgaud-claranet:
Formula 
`    signal = (metered/permitted.scale(60)).scale(100).publish('signal')
`
is incorrect.
PermittedThroughput need to be converted in second

https://docs.aws.amazon.com/efs/latest/ug/monitoring-metric-math.html#metric-math-throughput-utilization

### DIFF
--- a/modules/integration_aws-efs/conf/05-permitted-throughput.yaml
+++ b/modules/integration_aws-efs/conf/05-permitted-throughput.yaml
@@ -12,8 +12,7 @@ signals:
     metric: "PermittedThroughput"
     filter: "filter('stat', 'sum')"
   signal:
-    formula:
-      (metered/permitted.scale(1024)).scale(100)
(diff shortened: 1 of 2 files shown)


pdecat (approved):
LGTM :)

tchernomax (approved):
LGTM

tchernomax (changes requested):
You change detectors-gen.tf → you need to change [module/in](https://github.com/claranet/terraform-signalfx-detectors/blob/master/modules/integration_aws-efs/conf/05-permitted-throughput.yaml#L16) instead and run:
```
# to enter the docker container
make
# to generate the modules/integration_aws-efs/detectors-gen.tf from the conf
make update-module
```

pdecat (approved):

PR #418, opened 2022-06-13
AZ-578 Add fame_azure-vpn detectors

jmapro:
Create new detector for Azure VPN based on FAME metrics


modules/fame_azure-vpn/filters.tf
```diff
@@ -0,0 +1,5 @@
+locals {
```
BzSpi, line 1: The filters should be a link to https://github.com/claranet/terraform-signalfx-detectors/blob/master/common/module/filters-integration-azure.tf

BzSpi (changes requested):

Shr3ps (approved):

modules/fame_azure-vpn/filters.tf
jmapro: As discussed internally, I've reverted the link to filters-integration-azure. FAME act as an agent so it must use non prefixed tags.

modules/fame_azure-vpn/conf/01-total-flow-count.yaml
```diff
@@ -0,0 +1,21 @@
+module: Azure VPN
+name: Total Flow Count
+id: totalflowcount
+
+transformation: true
```
BzSpi, line 5: A transformation is needed here. I think that "max" is a good one.

BzSpi, line 5: A transformation is needed here. I think that "max" is a good one.

BzSpi (changes requested):

BzSpi (approved):

modules/fame_azure-vpn/conf/01-total-flow-count.yaml
```diff
@@ -0,0 +1,16 @@
...
+aggregation: true
+
+signals:
+  signal:
+    metric: fame.azure.virtual_network_gateway.total_flow_count
+
+rules:
+  critical:
+    threshold: 0
+    comparator: "=="
```
Shr3ps, line 15: open discussion: isn't it better "< 1"?

Shr3ps (approved):

modules/fame_azure-vpn/conf/01-total-flow-count.yaml
jmapro: no, because 0.5 is a valid value.

PR #430, opened 2022-07-21
Smart agent winprocess

nicolasvion:
Adding a way to monitor windows process throught signalfx.

Example SFX Agent configuration **telegraf/win_services section** : 
```yaml
    - type: telegraf/win_services
      serviceNames:
        - W3SVC
        - Tomcat8.5.11
```

### DIFF
--- a/modules/smart-agent_winprocess/conf/01-winprocess.yaml
+++ b/modules/smart-agent_winprocess/conf/01-winprocess.yaml
@@ -1,0 +1,15 @@
+module: winprocess 
+name: status
+
+transformation: ".mean(over='15m')"
+aggregation: true
+
+signals:
+  A:
+    metric: win_services.state
+  signal:
+    formula: A
+rules:
+  critical:
+    threshold: 4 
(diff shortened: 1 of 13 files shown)


modules/smart-agent_winprocess/conf/readme.yaml
```diff
@@ -0,0 +1,3 @@
+documentations:
```
pdecat, line 1: Please add links to documentation: https://github.com/signalfx/integrations/blob/master/signalfx-agent/agent_docs/monitors/telegraf-win_services.md

modules/smart-agent_winprocess/conf/01-winprocess.yaml
```diff
@@ -0,0 +1,15 @@
+module: winprocess
```
pdecat, line 1: Please name this detector `win-services` to match the underlying `win_services` metric that is used.

And maybe also rename the module `win-services if it is intended to only monitor Windows Services.

PR #434, opened 2022-08-24
Rework and harmonize HTTP 4xx detectors

BzSpi:


pdecat (approved):
Apart from the thresholds which I'd rather see at 90% for major, and 50 for minor, LGTM.

cvauvarin (approved):

PR #444, opened 2022-10-31
Add Azure CDN detector

Gati0:
Add the Azure CDN classic integration module.
This module aims to supervise the CDN resource of Azure (microsoft.cdn/profiles). For the moment only the latency is supervised.

### DIFF
--- a/modules/integration_azure-cdn/conf/00-latency.yaml
+++ b/modules/integration_azure-cdn/conf/00-latency.yaml
@@ -1,0 +1,17 @@
+module: Azure CDN
+name: latency
+filtering: "filter('resource_type', 'microsoft.cdn/profiles') and filter('primary_aggregation_type', 'true')"
+
+transformation: ".min(over='15m')"
+
+signals:
+  signal:
+    metric: "TotalLatency"
+rules:
+  critical:
+    threshold: 2000
+    comparator: ">"
+  major:
+    threshold: 4000
+    comparator: ">"
+    dependency: critical
--- a/modules/integration_azure-cdn/conf/00-latency.yaml
+++ b/modules/integration_azure-cdn/conf/00-latency.yaml
@@ -2,6 +2,7 @@ module: Azure CDN
 name: latency
 filtering: "filter('resource_type', 'microsoft.cdn/profiles') and filter('primary_aggregation_type', 'true')"
 id: cdn_total_latency
+value_unit: ms
 
 transformation: true
 aggregation: true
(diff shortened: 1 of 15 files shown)


modules/integration_azure-cdn/conf/00-latency.yaml
```diff
@@ -0,0 +1,24 @@
...
+aggregation: true
+
+signals:
+  signal:
+    metric: "TotalLatency"
+    rollup: max
+
+rules:
+  critical:
+    threshold: 1500
```
BzSpi, line 16: Are thresholds inverted ?

modules/integration_azure-cdn/conf/00-latency.yaml
```diff
@@ -0,0 +1,24 @@
...
+filtering: "filter('resource_type', 'microsoft.cdn/profiles') and filter('primary_aggregation_type', 'true')"
+id: cdn_total_latency
+
+transformation: true
+aggregation: true
+
+signals:
+  signal:
+    metric: "TotalLatency"
+    rollup: max
```
BzSpi, line 12: This should be removed, It would make the metric less readable in detector and also trigger more alerts than needed.

modules/integration_azure-cdn/conf/00-latency.yaml
```diff
@@ -0,0 +1,24 @@
+module: Azure CDN
+name: latency
+filtering: "filter('resource_type', 'microsoft.cdn/profiles') and filter('primary_aggregation_type', 'true')"
+id: cdn_total_latency
+
```
BzSpi, line 5: Add `value_unit` as described here https://github.com/claranet/terraform-signalfx-detectors/blob/master/scripts/templates/values.yaml#L99

modules/integration_azure-cdn/conf/readme.yaml
```diff
@@ -0,0 +1,3 @@
+documentations:
+
```
BzSpi, line 2: Please add a link to MS doc https://learn.microsoft.com/en-us/azure/azure-monitor/essentials/metrics-supported#microsoftcdnprofiles

BzSpi (changes requested):

modules/integration_azure-cdn/conf/readme.yaml
BzSpi: I don't see any change here :thinking:

Shr3ps (approved):

PR #446, opened 2022-11-15
Fix dependency rules

aureldsn:
- Some detectors have a wrong dependency rule (the dependency is applied on critical rule instead of major rule), so the critical rule is never triggered
- Add a dependency for the major rule of the response time detector in the smart-agent_http module


BzSpi (approved):

BzSpi:
Thanks for this great contribution !

PR #451, opened 2022-12-16
Change haproxy transformation : min to max for backend and server status

cvauvarin:
Change transformation function to not be alerted if a server or a backend is DOWN only once during the last 5 minutes.


Shr3ps (approved):

pdecat (approved):

tchernomax (approved):
LGTM

But IMHO we should get rid of this "transformation" and use "lasting_duration" and "lasting_at_least" (it would make the graph clearer)

https://github.com/claranet/terraform-signalfx-detectors/blob/58fa0e36b487aa15d32a2daf3239a35442c46b32/modules/smart-agent_redis/conf/07-memory-fragmentation-high.yaml#L20

PR #453, opened 2023-01-02
smart-agent_system-common: add swap in/out detector

tchernomax:
"memory" detector consider memory as a whole (RAM+swap) so it doesn't alert when there is a lot of memory pressure on hosts with a lot of swap compared to RAM (eg. 1G RAM , 0.5G swap).

This detector aim to fix this problem.

I set the default thresholds (10k and 50k) based on the values I am seeing on the incident I am working on ( ISS-427141 ) but I don't really know if they are good (maybe we will need some tweaking later).

I will try on my project ( PRJ-5PBQKN ).

note: the metric this new detector used need the [vmem](https://github.com/signalfx/signalfx-agent/blob/main/docs/monitors/vmem.md) monitor (which wasn't used by any detector in this repo before)

### DIFF
--- a/docs/severity.md
+++ b/docs/severity.md
@@ -1193,6 +1193,7 @@
 |System disk space utilization|X|X|-|-|-|
 |System disk inodes utilization|X|X|-|-|-|
 |System memory utilization|X|X|-|-|-|
+|System swap in/out|X|X|-|-|-|
 |System disk space running out|-|X|-|-|-|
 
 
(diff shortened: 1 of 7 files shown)


tchernomax:
If you want to see the detectors, there are deploy on PRJ-5PBQKN : https://git.fr.clara.net/CMP-CL585J/internals/PRJ-5PBQKN/terraform-monitoring/-/blob/master/signalfx-detectors/detector_swap.tf

PR #456, opened 2023-02-02
Fix Azure Redis Heartbeat aggregation

BzSpi:


pdecat (approved):

modules/integration_azure-redis/variables.tf
```diff
@@ -41,7 +41,7 @@ variable "heartbeat_timeframe" {
 variable "heartbeat_aggregation_function" {
   description = "Aggregation function and group by for heartbeat detector (i.e. \".mean(by=['host'])\")"
   type        = string
-  default     = ".mean(by=['shardid'])"
+  default     = ".mean(by=['azure_resource_id', 'shardid'])"
```
Shr3ps, line 44: Why not like others detectors?

```suggestion
  default     = ".mean(by=['shardid', 'azure_resource_name', 'azure_resource_group_name', 'azure_region'])"
```

Shr3ps (changes requested):

modules/integration_azure-redis/variables.tf
BzSpi: Because those dimensions do not exist :shrug:

Shr3ps: 😵‍💫

Shr3ps (approved):

modules/integration_azure-redis/variables.tf
BzSpi: Needs to be double check, dimensions are not displayed but seems available.

BzSpi: Changed

PR #461, opened 2023-03-15
Add splunk otel kubernetes common detectors

egouraud-claranet:
Due to deprecation of smart-agent, we need to propose a module that is splunk otel compatible for kubernetes.
Metrics and dimensions were renamed and the purpose of this merge request is to support them correctly to allow alerts to be raised properly.
as the directory was named with smart-agent, we propose to change it by splunk-otel prefix.
integrations should simply use this new module name to work.
source: https://docs.splunk.com/Observability/gdi/opentelemetry/legacy-otel-mappings.html#opentelemetry-values-and-their-legacy-equivalents

### DIFF
--- a/docs/severity.md
+++ b/docs/severity.md
@@ -69,6 +69,7 @@
 - [integration_gcp-pubsub-topic](#integration_gcp-pubsub-topic)
 - [integration_newrelic-apm](#integration_newrelic-apm)
 - [organization_usage](#organization_usage)
+- [otel-collector_kubernetes-common](#otel-collector_kubernetes-common)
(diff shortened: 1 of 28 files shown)


modules/splunk-otel_kubernetes-common/common-filters.tf
```diff
@@ -0,0 +1 @@
+../../common/module/filters-smart-agent.tf
```
haedri, line 1: someday we should rename this file.... out of scope for this PR I think, could be too complex

tmsdce (approved):
Ok de mon côté

PR #463, opened 2023-03-30
Rework Azure flexible PostgreSQL detectors

BzSpi:


modules/integration_azure-flexible-postgresql/conf/00-heartbeat.yaml
```diff
@@ -0,0 +1,15 @@
+---
+module: "Azure flexible PostgreSQL"
```
Shr3ps, line 2:
```suggestion
module: "Azure PostgreSQL flexible"
```

?

Shr3ps (approved):
We need the same change for MySQL Flexible

PR #466, opened 2023-04-06
Azure Backup detectors for VM & File Shares

BzSpi:


modules/integration_azure-backup/common-filters.tf
```diff
@@ -0,0 +1,4 @@
+locals {
```
Shr3ps, line 1: must be a symlink to common

Shr3ps (changes requested):

Shr3ps (approved):

alanbty (approved):

zfiel (approved):

BzSpi:
Back in Draft, behavior is not the one expected.

PR #467, opened 2023-04-06
Add Swap utilization detector

jmapro:


Poil:
Monitoring swap usage is useless if it's not correlated to memory usage, Kernel will always try to move low priority to Swap.
We should monitors pages in and out of swap space, and system free memory.
Also what to do when you trigger swap is full ? Disable swap and enable it again ?

Good article : https://support.cpanel.net/hc/en-us/articles/360060508193-My-server-is-continuously-using-large-amounts-of-swap-space

PR #470, opened 2023-04-17
Replace backend latency aggregation function sum by mean

BenjyOC:
It doesn't make sense to use `sum()` function for the bakend latency.  I replaced it by `mean()` function.

### DIFF
--- a/modules/integration_gcp-load-balancing/variables-gen.tf
+++ b/modules/integration_gcp-load-balancing/variables-gen.tf
@@ -212,7 +212,7 @@ variable "backend_latency_service_notifications" {
 variable "backend_latency_service_aggregation_function" {
   description = "Aggregation function and group by for backend_latency_service detector (i.e. \".mean(by=['host'])\")"
   type        = string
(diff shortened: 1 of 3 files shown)


modules/integration_gcp-load-balancing/variables-gen.tf
```diff
@@ -212,7 +212,7 @@ variable "backend_latency_service_notifications" {
 variable "backend_latency_service_aggregation_function" {
   description = "Aggregation function and group by for backend_latency_service detector (i.e. \".mean(by=['host'])\")"
   type        = string
-  default     = ".sum(by=['forwarding_rule_name', 'backend_target_name'])"
+  default     = ".mean(by=['forwarding_rule_name', 'backend_target_name'])"
```
pdecat, line 215: To also do in the configuration file used by templating: https://github.com/claranet/terraform-signalfx-detectors/blob/1c2cecad3d0820044086a245ea84323f5a35da0f/modules/integration_gcp-load-balancing/conf/03-latency-service.yaml#L6

JMHAUTBOIS-Claranet (approved):

PR #480, opened 2023-07-05
integration_aws-rds-common: fix

tchernomax:
MR created for Nicolas.

I will let him describe the problem and solution.

### DIFF
--- a/modules/integration_aws-rds-common/conf/05-dbload.yaml
+++ b/modules/integration_aws-rds-common/conf/05-dbload.yaml
@@ -4,19 +4,18 @@ name: "DB Load"
 
 transformation: true
 aggregation: true
-filtering: "filter('namespace', 'AWS/RDS')"
-value_unit: "%"
+filtering: "filter('namespace', 'AWS/RDS') and filter('stat', 'mean')"
 
 signals:
   signal:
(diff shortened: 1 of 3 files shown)


nicolas-dsk-claranet:
Currently, the metrics corresponding to dbload comes with the attribute 'stat'.
For a given instance, we will get five differents values for that stat attribute : mean, sum, count, lower, upper.
It's within the same detector, so the same thresold which doesnt make any sense then.
The pertinent one here is "mean", therefore the filter for that detector needs to be updated.

nicolas-dsk-claranet:
Indeed, i will change the default unit.
However, for the thresold, it will depend of the instance type and its model. 
What would be interesting to have as thresold ? I would propose Major 8 Crit 16

haedri:
I ran some calculations, in our case in one week the maximum was <1 for 20 orgs, 7 orgs between 1 and 8
And only one org had a max DBLoad of 15

So 8/16 seems good to me, to be adapted indeed depending on instance type/model

Note to ourselves : don't forget to include that in the changelog (and warn those concerned because who reads changelog anyway :) )

JMHAUTBOIS-Claranet (approved):

PR #481, opened 2023-07-12
Fix gcp memorystore threshold percent by ratio

BenjyOC:


haedri:
I wonder if it would be best to scale(100) instead of modifying the default value, maybe someone with more background has a suggestion ?
If not then default unit should be changed in detector & documentation

JMHAUTBOIS-Claranet (approved):

PR #482, opened 2023-07-25
Update CODEOWNERS

haedri:


.github/CODEOWNERS
```diff
@@ -1 +1 @@
-*   @bzspi @cvauvarin @jmhautbois-claranet @pdecat @shr3ps @tchernomax
+*   @bzspi @cvauvarin @jmhautbois-claranet @pdecat @shr3ps @tchernomax @haedri
```
pdecat, line 1: Please keep the list sorted.

pdecat: How about referencing teams here instead?

https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#codeowners-syntax

JMHAUTBOIS-Claranet: @pdecat yes, good idea. We have to solve some issues now, but we'll update this file aftewards.

JMHAUTBOIS-Claranet (approved):

PR #485, opened 2023-09-06
Fix smartagent postgree : do not compute replication lag for master node

lionel-kaufmann-claranet:
The replication lag is the delay beetween a replica ( stand_by ) and the master.
Compute a lag beetween a master and himself does not make any sense


pdecat (approved):
LGTM!

For reference: https://docs.splunk.com/Observability/gdi/monitors-databases/postgresql.html

PR #486, opened 2023-09-13
try to avoid alert flaps..

lionel-kaufmann-claranet:
Splunk otel agent use supervisor's RPC interface to collect status.
That's why in https://docs.splunk.com/Observability/gdi/monitors-hosts/supervisor.html for supervisor.state metric they directly link the gauge values to supervisor codes : http://supervisord.org/subprocess.html#process-states

Now consider a particular supervisor code 10 : Starting.

Stage 1:  the PRODUCTION Way 
"What ? oh ! this is legacy code, Well, it was developed by a guy who left company 12 years ago.... wait...who was the guy ? Did not remember...But no matter the guy ! 
Do do not worry ! We have solve stability issues by restarting all this process every hours..." 
next ...

Stage 2 : the DEVEL Way
"What ? oh ! this is legacy code, with out-of-life libraries, with out-of-life design patterns, with npe, with home made libraries, with memory leaks and sometime a lock blocking condition...did i forget something ? 
Do do not worry ! We have solve stability issues by restarting all this process every hours..."
next...

Stage 3: the BUDGET Way ( endly )
 "What ? oh ! this is  legacy code, there is an incredible tech debt there ! Refactoring and rebuild price is very high with no added value...Therefore  the line in the budget approval list is top bottom.  
Do do not worry ! We have solve stability issues by restarting all this process every hours..." 

Frequency : ~ 24 times per day
It's a bit noisy to have major alert there...
^_^

By the way... all this noise for a RTFM : 
- stop code is 0 
- starting code is 10... this is definitely less than 20 !

Should we consider removing var.process_state_threshold_major ?


lionel-kaufmann-claranet:
Did not find "not ready" button on github... wait for merge...

modules/smart-agent_supervisor/detectors-supervisor.tf
```diff
@@ -36,7 +36,7 @@ resource "signalfx_detector" "process_state" {
   program_text = <<-EOF
     signal = data('supervisor.state', filter=${module.filtering.signalflow})${var.process_state_aggregation_function}${var.process_state_transformation_function}.publish('signal')
     detect(when(signal > ${var.process_state_threshold_critical})).publish('CRIT')
-    detect(when(signal < ${var.process_state_threshold_major})).publish('MAJOR')
+    detect(when(signal == 0 ).publish('MAJOR')
```
pdecat, line 39: I'd rather change default value of the `process_state_threshold_major` variable.

lionel-kaufmann-claranet: But a threshold_major of 10 would involve only one possible code : 0 = stopped.
This let some possibility in the codebase to implement subsequent  return code < 10,  if one day superversor change his 10*code pattern design for return codes... 
Agreed. -> Pushed

pdecat (approved):
LGTM!

PR #487, opened 2023-09-21
fix redis keyspace_hits, use derive.keyspace_hits

vsananikone:
The hitrate detector doesn't have any metrics cause `counter.keyspace_hits` does not exist.
the metric used in collectd/redis for signal-fx is `derive.keyspace_hits`
https://github.com/signalfx/signalfx-agent/blob/main/docs/monitors/collectd-redis.md

and our collector splunk-otel-collector is using the correct metric `derive.keyspace_hits`


pdecat (approved):
LGTM!

PR #499, opened 2023-10-03
Rework Azure SQL Database

BzSpi:


modules/integration_azure-sql-elastic-pool/conf/02-storage-usage.yaml
```diff
@@ -0,0 +1,21 @@
+---
+module: "Azure SQL Database"
+name: "storage usage"
+filtering: "filter('resource_type', 'Microsoft.Sql/servers/databases') and filter('primary_aggregation_type', 'true')"
```
haedri, line 4: Shouldn't it be elasticpools instead of databases ?

modules/integration_azure-sql-elastic-pool/conf/03-dtu-consumption.yaml
```diff
@@ -0,0 +1,21 @@
+---
+module: "Azure SQL Database"
+name: "DTU consumption"
+filtering: "filter('resource_type', 'Microsoft.Sql/servers/databases') and filter('primary_aggregation_type', 'true')"
```
haedri, line 4: Shouldn't it be elasticpools instead of databases ?

haedri (changes requested):

modules/integration_azure-sql-elastic-pool/conf/02-storage-usage.yaml
BzSpi: Good catch !
Fixed.

modules/integration_azure-sql-elastic-pool/conf/03-dtu-consumption.yaml
BzSpi: Good catch !
Fixed.

haedri (approved):

PR #506, opened 2023-10-17
fix : correct invalid sed interval

haedri:
One generator sed command was failling on Ubuntu 22.04 because of invalid interval a-zA-z instead of a-zA-Z


pdecat (approved):
LGTM!

Shr3ps (approved):

PR #507, opened 2023-10-19
Wrong hitrate calculation for smart agent Redis 

aogereau:
Currently, the Redis hitrate detector for smart_agent_redis uses the following formula

A / ( A + B )

Where both A and B have the same value ( derive.keyspace_hits ) which will always give a result of 50 % .

this fix will replace B by derive.keyspace_misses instead,  (according to the official documentation : https://github.com/signalfx/signalfx-agent/blob/main/docs/monitors/collectd-redis.md ) , which will give the correct value for the redis hitrate

image


pdecat (approved):
LGTM!

PR #508, opened 2023-10-19
fix(aws-backup): avoid duplicated and ill-routed alerts

NSenaud:


modules/integration_aws-backup/README.md
```diff
@@ -57,7 +58,7 @@ Note the following parameters:
 
 These 3 parameters alongs with all variables defined in [common-variables.tf](common-variables.tf) are common to all
```
pdecat, line 59:
```suggestion
These 4 parameters along with all variables defined in [common-variables.tf](common-variables.tf) are common to all
```

pdecat: Never mind, I did not see  the folded section.

Also, the `alongs` typo is everywhere...

NSenaud: > Note the following parameters:
> * `source`: [...]
> * `environment`: [...]
> * `notifications`: [...]
> These 3 parameters [...] are common to all modules in this repository. Other variables, specific to this module, are available in variables.tf [...].

Looks good to me, since the "3" does not refer to the code example but the list just above. The source template needs to be fixed however, doing another PR for the typo since it will affect every stacks.

pdecat: Yes, that's why I said _Never mind_ :)

NSenaud: https://github.com/claranet/terraform-signalfx-detectors/pull/510

NSenaud: Github UI lagging, I wasn't seeing your comment, until I reload the page :sweat_smile:

modules/integration_aws-backup/variables.tf
```diff
@@ -0,0 +1,3 @@
+variable "aws_account_id" {
```
haedri, line 1: I don't think forcing aws_account_id is a good idea, not everyone wants to filter by aws_account_id
It's best to let people choose if they want to by setting the appropriate filter in their own terraform

modules/integration_aws-backup/common-filters.tf
```diff
@@ -1,3 +1,11 @@
...
+#
+# We are also assuming that we want to be notified for every backup vault and
+# resource types. `match_missing=False` is required to avoid duplicated signal
+# and alerts, since CloudWatch returns aggregated metric, and per vault and
+# resource type.
 locals {
-  filters = "filter('ResourceType', '*')"
-}
\ No newline at end of file
+  filters = format("filter('namespace', 'AWS/Backup') and filter('stat', 'sum') and filter('BackupVaultName', '*', match_missing=False) and filter('ResourceType', '*', match_missing=False) and filter('aws_account_id', '%s')", var.aws_account_id)
```
haedri, line 10: Aren't signals already filtered with most of these variables in the detector conf ?

modules/integration_aws-backup/variables.tf
NSenaud: Hello, we don't have much choice because recovery points are not automatically tagged according to the convention used with the detectors (see comment in `common-filters.tf`).

modules/integration_aws-backup/common-filters.tf
NSenaud: I have removed the `filtering` field in the detectors conf :+1:

modules/integration_aws-backup/variables.tf
NSenaud: Variable removed after further discussions.

haedri (approved):

PR #509, opened 2023-10-20
Fix rabbitMQ monitoring

aogereau:
Two fixes : 

Fixed RabbitMQ Queue consumer use : correct name of the metric is `gauge.queue.consumer_utilisation` , and not `gauge.queue.consumer_use`

image

Added the metric `gauge.queue.messages` in the extra metrics for the otel configuration . According to the official documentation it is a custom metric : https://docs.splunk.com/observability/en/gdi/monitors-messaging/rabbitmq.html


pdecat (approved):
LGTM!

PR #512, opened 2023-11-04
prometheus-exporter_couchdb

hlepesant:
This detector is based on the prometheus embedded service in CouchDB [Configuration of Prometheus Endpoint](https://docs.couchdb.org/en/stable/config/misc.html#configuration-of-prometheus-endpoint)


haedri:
It seems fork are not supported, could you create a branch instead ? and update severity.md also

PR #515, opened 2023-11-13
fix #514 - wrong namespace

### LINKED ISSUE #514: [BUG] Detector backend_latency - module integration_aws-elb - Latency not retrieved
**What is the module?**
integration_aws-elb

**What is the detector?**
backend_latency

**Describe the bug**
The detector is currently looking to get the metric Latency from Cloudwatch with the filter for the namespace on "AWS/ApplicationELB" rather than "AWS/ELB". I put the documentation below.

ALB
https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-cloudwatch-metrics.html
ELB
https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/elb-cloudwatch-metrics.html

**Expected behavior**
Data shall be provided to the detector

nicolas-dsk-claranet:
Fix to the issue https://github.com/claranet/terraform-signalfx-detectors/issues/514

Check the issue, the solution is quite clear

### DIFF
--- a/modules/integration_aws-elb/conf/01-latency.yaml
+++ b/modules/integration_aws-elb/conf/01-latency.yaml
@@ -3,7 +3,7 @@ name: backend latency
 
 transformation: true
 aggregation: true
-filtering: "filter('namespace', 'AWS/ApplicationELB')"
+filtering: "filter('namespace', 'AWS/ELB')"
 value_unit: "Second"
 
 signals:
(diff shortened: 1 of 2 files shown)


haedri (approved):
Indeed no data for this detector on all our orgs... except 1 where the detector was corrected manually ;)

pdecat (approved):
Good catch!

PR #519, opened 2023-11-15
ignore squashfs type for inodes

aogereau:
Currently both inodes detectors on module smart-agent_system-common can detect squashfs partitions , triggering false positive alerts at 100% utilisation .

System disk inodes utilization :

image

System filesystem inodes utilization :  

image

This fix will add a filter on squashfs partitions for both detectors , to prevent false positive alerts

### DIFF
--- a/modules/smart-agent_system-common/conf/04-filesystem-inodes.yaml
+++ b/modules/smart-agent_system-common/conf/04-filesystem-inodes.yaml
@@ -6,10 +6,10 @@ value_unit: "%"
 signals:
   used:
     metric: system.filesystem.inodes.usage
-    filter: filter('state', 'used')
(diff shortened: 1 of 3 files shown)


modules/smart-agent_system-common/conf/04-filesystem-inodes.yaml
```diff
@@ -6,10 +6,10 @@ value_unit: "%"
 signals:
   used:
     metric: system.filesystem.inodes.usage
-    filter: filter('state', 'used')
+    filter: "(filter('state', 'used') and not filter('type', 'squashfs'))"
```
pdecat, line 9: This sort of filtering is supposed to happen at the collection level (smartagent or otel-collector).

And `squashfs` is already excluded by default.

How are you collecting the metrics?
Do you have a custom configuration?

PR #522, opened 2023-11-16
fix: use lasting instead of transformation function

Tulux:
Similarly to efcee1ff0aeb397850aed3a8cdd053e3e55e975a, replace transformation function with appropriate `when()` **lasting** argument for various system detectors.


pdecat (approved):
LGTM!

haedri:
For the release note:
Detectors "disk space utilization", "filesystem inodes utilization", and "disk inodes utilization" will now trigger after lasting duration (5m by default) and not immediately.
You can set transformation to ".max(over='5m')" and removing the lasting duration to revert to old behavior

PR #542, opened 2024-01-10
Rework smart agent elasticsearch

Gati0:
Migration of all elasticsearch detectors to generated mode with the gen_detectors module.


modules/smart-agent_elasticsearch/conf/01-cluster-status.yaml
```diff
@@ -0,0 +1,19 @@
+module: ElasticSearch
+name: "cluster status"
+aggregation: true
```
BzSpi, line 3: If you have multiple clusters, this will do the transformation on all clusters :smiley:

modules/smart-agent_elasticsearch/conf/01-cluster-status.yaml
```diff
@@ -0,0 +1,19 @@
...
+transformation: ".mean(over='5m')"
```
BzSpi, line 4: Why do a mean on 5 minutes since thresholds are strict comparison and lasting is already set ?

modules/smart-agent_elasticsearch/conf/02-cluster_initializing_shards.yaml
```diff
@@ -0,0 +1,21 @@
+module: ElasticSearch
+name: "cluster initializing shards"
+aggregation: true
```
BzSpi, line 3: If you have multiple clusters, this will do the transformation on all clusters 😃

modules/smart-agent_elasticsearch/conf/02-cluster_initializing_shards.yaml
```diff
@@ -0,0 +1,21 @@
...
+transformation: ".min(over='15m')"
```
BzSpi, line 4: Already managed by triggering rule

BzSpi (changes requested):

PR #544, opened 2024-01-17
smart-agent_elasticsearch: use "extrapolation" & remove "fill"

tchernomax:
and also use "lasting" and "at_least" "when()" arguments


BzSpi:
Will be supersede by complete rewrite done here: https://github.com/claranet/terraform-signalfx-detectors/pull/542

PR #546, opened 2024-01-25
Aws ecs transformation fix

Tulux:
Fix AWS ECS monitor :

1. Remove usage of mean() which is dangerous to use as it may merge metrics of similar service names from different clusters (ex: cluster1/myservice and cluster2/myservice would be calculated as a single myservice metric)
2. Replace usage of min() transformation function with lasting() argument, similarly to #520 and #522

### DIFF
--- a/modules/integration_aws-ecs-service/detectors-ecs-service.tf
+++ b/modules/integration_aws-ecs-service/detectors-ecs-service.tf
@@ -35,7 +35,7 @@ resource "signalfx_detector" "cpu_utilization" {
   tags                    = compact(concat(local.common_tags, local.tags, var.extra_tags))
 
   program_text = <<-EOF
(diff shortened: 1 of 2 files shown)


modules/integration_aws-ecs-service/detectors-ecs-service.tf
```diff
@@ -35,9 +35,9 @@ resource "signalfx_detector" "cpu_utilization" {
   tags                    = compact(concat(local.common_tags, local.tags, var.extra_tags))
 
   program_text = <<-EOF
-    signal = data('CPUUtilization', filter=filter('namespace', 'AWS/ECS') and filter('stat', 'mean') and filter('ServiceName', '*') and ${module.filtering.signalflow}).mean(by=['ServiceName'])${var.cpu_utilization_aggregation_function}${var.cpu_utilization_transformation_function}.publish('signal')
```
pdecat, line 38: How about adding `ClusterName` to `mean` instead?

pdecat: Never mind.

pdecat (approved):
LGTM!

PR #550, opened 2024-04-02
Add a duration for which conditions must be true before triggering an alert on Integration_aws-rds-common

pierreislande:
Add a duration for which conditions must be true before triggering an alert on Integration_aws-rds-common
The default *lasting_duration* is set to none to avoid changing the actual behavior

### DIFF
--- a/modules/integration_aws-rds-common/detectors-rds-common.tf
+++ b/modules/integration_aws-rds-common/detectors-rds-common.tf
@@ -35,8 +35,8 @@ resource "signalfx_detector" "cpu_90_15min" {
 
   program_text = <<-EOF
     signal = data('CPUUtilization', filter=filter('namespace', 'AWS/RDS') and filter('stat', 'mean') and filter('DBInstanceIdentifier', '*') and ${module.filtering.signalflow})${var.cpu_90_15min_aggregation_function}${var.cpu_90_15min_transformation_function}.publish('signal')
-    detect(when(signal > ${var.cpu_90_15min_threshold_critical})).publish('CRIT')
-    detect(when(signal > ${var.cpu_90_15min_threshold_major}) and (not when(signal > ${var.cpu_90_15min_threshold_critical}))).publish('MAJOR')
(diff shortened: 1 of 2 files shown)


modules/integration_aws-rds-common/detectors-rds-common.tf
```diff
@@ -35,8 +35,8 @@ resource "signalfx_detector" "cpu_90_15min" {
 
   program_text = <<-EOF
     signal = data('CPUUtilization', filter=filter('namespace', 'AWS/RDS') and filter('stat', 'mean') and filter('DBInstanceIdentifier', '*') and ${module.filtering.signalflow})${var.cpu_90_15min_aggregation_function}${var.cpu_90_15min_transformation_function}.publish('signal')
-    detect(when(signal > ${var.cpu_90_15min_threshold_critical})).publish('CRIT')
-    detect(when(signal > ${var.cpu_90_15min_threshold_major}) and (not when(signal > ${var.cpu_90_15min_threshold_critical}))).publish('MAJOR')
+    detect(when(signal > ${var.cpu_90_15min_threshold_critical}, lasting=%{if var.cpu_90_15min_lasting_duration_critical == null}None%{else}'${var.cpu_90_15min_lasting_duration_critical}'%{endif}, at_least=${var.cpu_90_15min_at_least_percentage_critical})).publish('CRIT')
```
pdecat, line 38: Is that Jinja2 templating? I do not believe it is supported here.

pierreislande: Yeah, like here jinja is already used: https://github.com/claranet/terraform-signalfx-detectors/blob/master/modules/integration_aws-rds-common/detectors-gen.tf#L11
already tested and is ok.

pdecat: Oh, this is not Jinja2 templating but [Terraform templating](https://developer.hashicorp.com/terraform/language/expressions/strings#directives). Never used that all that time :facepalm:

pdecat (approved):
LGTM!

Note: I believe the wording of descriptions could be improved, but at least it is the same all over the place :)

PR #559, opened 2024-05-28
SNAT ports: add resource name in mean

freezeeedos:
Hi, 

azure_resource_name is used by some monitoring tools to populate useful fields used by run teams. I simply added this field to the mean function

### DIFF
--- a/modules/integration_azure-firewall/conf/01-snat.yaml
+++ b/modules/integration_azure-firewall/conf/01-snat.yaml
@@ -3,7 +3,7 @@ module: Azure firewall
 name: "SNAT Port Utilization"
 
 transformation: true
-aggregation: ".mean(by=['protocol', 'azure_resource_name'])"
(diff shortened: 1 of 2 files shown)


modules/integration_azure-firewall/conf/01-snat.yaml
```diff
@@ -3,7 +3,7 @@ module: Azure firewall
 name: "SNAT Port Utilization"
 
 transformation: true
-aggregation: ".mean(by=['protocol'])"
+aggregation: ".mean(by=['protocol', 'azure_resource_name'])"
```
BzSpi, line 6: We commonly use this format

```suggestion
aggregation: ".mean(by=['protocol', 'azure_resource_name', 'azure_resource_group_name', 'azure_region'])"
```

BzSpi (approved):

modules/integration_azure-firewall/conf/01-snat.yaml
freezeeedos: Thanks, it's fixed

PR #569, opened 2024-09-13
Fix gcp memory utilization

eclisson-clt:
Update/Fix du detector sur l'utilisation mémoire


haedri:
I think it can be problematic since thresholds are not the same at all (usage % vs free %), for example if previous variable `memory_utilization_threshold_critical` was manually set, then updating will trigger this detector a lot

2 solutions :

1. Change signal from free % to usage % (`100 - signal`), impacts will be minimal for users
2. Change filename memory_utilization so that variable name changes also

Solution 2 might be easier I think

PR #576, opened 2024-11-20
Fix nginx heartbeat

ftremblot:
Since the [v0.112](https://github.com/signalfx/splunk-otel-collector/releases/tag/v0.112.0) tag the collectd/nginx monitor is deprecated.

The goal of this PR is to support the use of the new [nginx receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/nginxreceiver/) 

I've deleted the "dropped_connections" detector since the new receiver doesn't have any of kind of replacement for `connections.failed`


haedri:
Generated files should not be edited manually, instead change / delete files in the conf/ directory of the module, and regenerate the terraform files afterwards (https://github.com/claranet/terraform-signalfx-detectors/blob/master/docs/development.md#detectors)

PR #577, opened 2024-12-09
Integration azure sql managed instances

rchauvetclara:
Add new detector for Azure SQL Managed Instances.

Two metrics are used : 

`avg_cpu_percent`
`storage_space_used_mb`

Tests OK. Validate on the cityzmedia customer monitoring configuration

### DIFF
--- a/modules/integration_azure-sql-managed-instances/conf/02-storage-usage.yaml
+++ b/modules/integration_azure-sql-managed-instances/conf/02-storage-usage.yaml
@@ -1,0 +1,21 @@
+---
+module: "Azure SQL Managed Instances"
+name: "storage usage"
+filtering: "filter('resource_type', 'Microsoft.Sql/managedInstances') and filter('primary_aggregation_type', 'true')"
+aggregation: ".mean(by=['azure_resource_name', 'azure_resource_group_name', 'azure_region'])"
+value_unit: "%"
+transformation: true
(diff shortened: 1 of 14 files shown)


modules/integration_azure-sql-managed-instances/conf/02-storage-usage.yaml
```diff
@@ -0,0 +1,17 @@
+---
+module: "Azure SQL Managed Instances"
+name: "storage usage"
+filtering: "filter('resource_type', 'Microsoft.Sql/managedInstances') and filter('primary_aggregation_type', 'true')"
+aggregation: ".mean(by=['azure_resource_name', 'azure_resource_group_name', 'azure_region'])"
+value_unit: "Megabyte"
+transformation: true
+signals:
+  signal:
+    metric: "storage_space_used_mb"
```
BzSpi, line 10: Must be compared to `reserved_storage_mb` metric to monitor the percentage instead of fixed value.

BzSpi (changes requested):

modules/integration_azure-sql-managed-instances/conf/02-storage-usage.yaml
rchauvetclara: Thanks for the suggestion. It's better now :)

Issue #587, opened 2025-04-17
[IMPROVE] Detector AmazonMQ RabbitMQ


nicolas-dsk-claranet:
**What is the module?**
[Please give the module name or link where detectors belongs to]
(https://github.com/claranet/terraform-signalfx-detectors/tree/master/modules/integration_aws-amazonmq-rabbitmq).

**What is the detector?**
All detectors from that module

**Is your improvement related to a known monitoring limitations? Please describe.**
The filter on aws_account_id is mandatory and we can't disable it

**Describe the solution you'd like**
An option to disable it or remove it directly (filtering_custom should be used instead imho)

**Describe alternatives you've considered**
Nope

**Additional context**
@tchernomax @lionel-kaufmann-claranet Coucou

PR #588, opened 2025-04-17
feat(aws_mq): remove filter deprecated

nicolas-dsk-claranet:


nicolas-dsk-claranet:
Implementation of a workaround for the issue #587

lionel-kaufmann-claranet:
- removed unused file
- link filter to common aws one
- Readme review

lionel-kaufmann-claranet:
There was a mistake in @nicolas-dsk-claranet diagnostic.
In the namespace amazonMQ we have the data of all the aws accounts, we definitly need to filter out the data for the environement, however the original mistake was that there is no collumn aws_account_id.

The filtering must be done on AWSUniqueId field. This field have a "special" format arn:aws:mq:<cluster_region>:<aws_account_id>:<cluster_name>
Put this format in the Readme file in order people to quickly understand how to fill this variable 
Note that you find inside it the aws_account_id, that previous version focused on, but this was not enough.

Deployed on 3 environment, and working now, do not hesitate to contact Nicolas or me to give you the account for validation

Reviewers : do not undestand why tests are falling, as the content does not feat modules, I suspect a border effect due to tag change ( see files changed for the typo fix )

